PR #15, opened 2022-05-06
test: MOPLAT-236 add tests

patrinoua:
This PR resolves [MOPLAT-236]

### Description

Adds a number of configuration options for testing:
- babel.config.js
- jest.config.js
- package.json*
- setupJest.ts

## TODO 
- remove schema and track.js

* **Question**: Should we avoid installing enzyme at all? It is currently used in a lot of eigen components which leaves us with two options: One copy paste the components and their tests or refactor the tests to not use enzyme every time we want to use a component 

### PR Checklist

### To the reviewers 👀

<details><summary>Changelog updates</summary>

#### Cross-platform user-facing changes

-

#### iOS user-facing changes

-

#### Android user-facing changes

-

#### Dev changes

-

</details>

Need help with something? Have a look on our [docs], or get in touch with us.

[docs]: /docs/README.md

[MOPLAT-236]: https://artsyproduct.atlassian.net/browse/MOPLAT-236?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


package.json
```diff
@@ -8,22 +8,32 @@
...
     "@react-navigation/native-stack": "^6.2.5",
     "@react-navigation/stack": "^6.0.11",
+    "@segment/analytics-react-native": "^2.1.12",
+    "@sentry/react-native": "^3.4.2",
+    "@storybook/addon-actions": "^6.4.22",
+    "@storybook/react-native": "^5.3.25",
     "@styled-system/theme-get": "^5.1.2",
+    "@testing-library/react-native": "^9.1.0",
     "easy-peasy": "^5.0.4",
+    "enzyme": "^3.11.0",
```
patrinoua, line 36: maybe we should add a danger rule like we have in [eigen](https://github.com/artsy/eigen/blob/main/dangerfile.ts#L36-L52) to avoid using it for newer test.s

src/setupJest.ts
```diff
@@ -0,0 +1,157 @@
...
+//       { countryCode: "FR", languageTag: "fr-FR", languageCode: "fr", isRTL: false },
+//     ]
+//   },
+//   getCurrencies() {
+//     return ["USD", "EUR"]
+//   },
+//   getTimeZone() {
+//     return "America/New_York"
+//   },
+// }))
```
rajsam003, line 66: So many lines of code are commented. Maybe we can remove it if it's not useful.

admbtlr:
This looks like at least three different PRs merged into one (dependency updates, storybook, adding tracking, ...). Could you break it out into separate PRs please?

babel.config.js
```diff
@@ -1,32 +1,66 @@
...
-  'react-native-reanimated/plugin',
-]
-
 if (process.env.CI) {
   // When running a bundled app, these statements can cause a big bottleneck in the JavaScript thread.
   // This includes calls from debugging libraries and logs we might forget to remove
   plugins.push("transform-remove-console")
 }
 
-module.exports = { presets, plugins }
```
MounirDhahri, line 32: **thought:** any reason to separate between those?
**issue:** the lines from 1 to 5 will have no meaning and the optimization for removing the console logs is no longer going to be applied since there is no plugins object

babel.config.js
```diff
@@ -1,32 +1,66 @@
...
+      "@babel/plugin-transform-runtime",
+      [
+        "@babel/plugin-proposal-decorators",
+        {
+          legacy: true, // this is only needed for `ProvideScreenTracking` that is deprecated. once we dont have that anymore, we can remove this. probably the whole plugin actually.
+        },
+      ],
+      ["@babel/plugin-proposal-class-properties", { loose: true }],
+      ["@babel/plugin-proposal-private-methods", { loose: true }],
+      ["@babel/plugin-transform-react-jsx", { runtime: "automatic" }], // this is so `import React from "react"` is not needed.
```
MounirDhahri, line 45: **issue:** lots of those plugins are already included in the react native babel preset, are they all needed? also I don't feel like it's a good idea to already set them to `loose`. in eigen that was fine because of the number of dependencies there, I don't think however we should do that. It will make maintaining this really hard in the future, I would recommend sticking to what we get from the preset and only include plugins that we really need for the tests

package.json
```diff
@@ -8,22 +8,32 @@
...
-    "@react-native-async-storage/async-storage": "^1.15.14",
+    "@expo/react-native-action-sheet": "^3.13.0",
+    "@react-native-async-storage/async-storage": "^1.17.4",
+    "@react-native-community/clipboard": "^1.5.1",
     "@react-navigation/bottom-tabs": "^6.0.9",
     "@react-navigation/native": "^6.0.6",
     "@react-navigation/native-stack": "^6.2.5",
     "@react-navigation/stack": "^6.0.11",
+    "@segment/analytics-react-native": "^2.1.12",
+    "@sentry/react-native": "^3.4.2",
```
MounirDhahri, line 30: **thought:** I am lost, I thought the PR is to add tests and not tracking

package.json
MounirDhahri: **issue:** why do we need to add it actually? we can just not use it already

babel.config.js
```diff
@@ -1,32 +1,66 @@
...
+        "module-resolver",
+        {
+          root: ["./"],
+          alias: {
+            "@Scenes": "./src/Scenes",
+            "@store": "./src/store",
+            "@relay": "./src/relay",
+            "@routes": "./src/routes",
+            "@helpers": "./src/helpers",
+            "@assets": "./src/assets",
```
damassi, line 58: `"foo": "./src/foo"` still seems a bit nicer than having to add a special character (which isn't necessary)

src/utils/track/index.tsx
```diff
@@ -0,0 +1,233 @@
...
+}
+
+interface ProvideScreenTrackingProps {
+  info: Schema.PageView
+}
+
+/** Deprecated
+ * Please use `ProvideScreenTrackingWithCohesionSchema` instead.
+ */
+@screenTrack<ProvideScreenTrackingProps>((props) => props.info)
```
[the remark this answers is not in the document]
damassi: (We also don't need introduce classes; everything should be converted into an FC.)

babel.config.js
patrinoua: @MounirDhahri They are mainly copy paste. Where can I see which plugins are already preinstalled to remove them from here?

package.json
```diff
@@ -8,22 +8,32 @@
...
     "lint": "eslint . --ext .js,.jsx,.ts,.tsx",
-    "setup:artsy": "./scripts/setup-env-for-artsy",
     "install:all": "yarn install",
     "pod-install": "cd ios && bundle exec pod install && cd ..",
     "postinstall": "patch-package",
+    "relay": "yarn run relay-compiler --watch",
+    "setup:artsy": "./scripts/setup-env-for-artsy",
     "sync-schema": "curl https://raw.githubusercontent.com/artsy/metaphysics/main/_schemaV2.graphql -o data/schema.graphql; yarn prettier --write --parser graphql data/schema.graphql",
-    "relay": "yarn run relay-compiler --watch"
+    "type-check": "yarn relay; yarn tsc"
```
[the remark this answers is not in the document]
patrinoua: copy paste from eigen! :-)

patrinoua:
replaced by https://github.com/artsy/energy/pull/21

PR #22, opened 2022-05-12
feat: add color to energy repo

patrinoua:
### Description

It often happens that I think I'm working on energy but working on eigen instead so I thought I'd add this to help us separate between repos.

1) I know Sam likes this idea but it has not been broadly discussed so if it's not welcome of course we can revert this.
1) If we decide to close it, I am not sure how to have this locally, does anyone know?

<details><summary>Changelog updates</summary>

#### Dev changes
add color to VSCode tab


pvinis:
lets keep these on user config, not repo config.

PR #31, opened 2022-05-16
test: speed up tests. clean up some test setup.

pvinis:
This PR resolves []

### Description

⚠️ Please run it on your machines too to confirm the "after" numbers are close to mine. `yarn install`, then `yarn test` should do it. and then run `yarn test` again for a cache variation.

Before:
- 12 sec with cache
- 55 without cache

After:
- 6 with cache
- 20 without cache

Not perfect, but better. I noticed that for our 2 test files, each takes about 2-3 seconds to start, so maybe something in our setupJest is taking long? I tried to even keep that file empty, and still that delay was there, which confused me.

In any case, the tests are a bit faster now, and we can take a look later again.

(I tried vitest quickly, but apparently RN is not yet supported. Hopefully some day soon.)

### PR Checklist

### To the reviewers 👀

[docs]: /docs/README.md


pvinis:
oh, i also removed the "renders without error" test. its not really a good test, especially with relay not resolving. we basically test that an empty screen will render.

pvinis:
oh god, the conflicts haha. ill fix now.

rajsam003 (approved):
This looks good to me.
Thanks for this PR @pvinis 🙏🏽

src/setupJest.ts
```diff
@@ -120,18 +32,7 @@ jest.mock("react-native-device-info", () => ({
   hasNotch: jest.fn(),
 }))
 
-const _ = jest.requireActual("lodash")
-
-jest.mock("react-native-reanimated", () => require("react-native-reanimated/mock"))
-
-jest.mock("react-native/Libraries/LayoutAnimation/LayoutAnimation", () => ({
```
rajsam003, line 127: Why dont we need any of these jest setups?

src/Screens/HomeTabs/HomeTabs.tests.tsx
```diff
@@ -9,17 +8,14 @@ jest.unmock("react-relay")
...
+describe("HomeTabs", () => {
+  const TestRenderer = () => (
+    <RelayEnvironmentProvider environment={mockEnvironment}>
+      <HomeTabs />
+    </RelayEnvironmentProvider>
+  )
 
   it("renders without throwing an error", () => {
-    getWrapper()
+    //     renderWithWrappersTL(<TestRenderer />)
```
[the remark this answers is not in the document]
pvinis: I wrote a comment in this PR. it's not a good or useful test, so I used it for checking the time, but then commented it out

src/setupJest.ts
pvinis: it seems we don't need them yet. let's add them once we do need them.

rajsam003:
Lemme run this branch on my end before merging it.

rajsam003:
I am getting this error on my simulator. Do I have to run any command? 

image

pvinis:
thanks for trying it. huh. I guess I fixed the type but didnt assign it?

if you're already there, can you add the first line from https://github.com/artsy/eigen/blob/main/index.ios.js to the index.js file and see if that fixes this?
andnifnit works you can commit it.

patrinoua (approved):

PR #39, opened 2022-05-23
refactor: Removing test for all tabs 

rajsam003:
This PR resolves []

### PR Checklist

### To the reviewers 👀

[docs]: /docs/README.md


src/Screens/HomeTabs/Artists/ArtistTabs/ArtistTabs.tsx
```diff
@@ -31,8 +32,12 @@ const Header = ({ artistName, navigation }: TabBarProps & HeaderProps) => {
...
 }
 
-export const ArtistTabs = ({ route, navigation }: ArtistTabsProps) => {
+export const ArtistTabs = () => {
   const insets = useSafeAreaInsets()
+
+  const route = useRoute<ArtistTabsProps["route"]>()
+  const navigation = useNavigation<ArtistTabsProps["navigation"]>()
+
   const { artistName } = route.params
```
pvinis, line 41: are we doing slug here?

rajsam003: I tried with slug and with that, we use the query to get just the artist name to display as the header. So for that, we need to `useLazyLoadQuery` and `Suspense` which I think is a lot of work. So I thought it simple to just pass `artistName` in a `route.param`. What do you think?

rajsam003: Also I am getting this annoying issue even though I am using `<Suspense fallback={<ActivityIndicator />}>`

image

rajsam003: https://stackoverflow.com/questions/54432861/a-react-component-suspended-while-rendering-but-no-fallback-ui-was-specified

pvinis: lets merge for now, and we can figure this out afterwards.

for sure we need to pass the slug, so we can actually do the relay query, but lets not block this PR.

rajsam003: @MounirDhahri helped me with the solution to use slug 🙏🏽. I will use it on my next PR.

PR #45, opened 2022-05-30
feat: blue icon!

admbtlr:
This PR resolves []

### Description

Same as the old Folio icon, but using Artsy Blue instead of Artsy Purple. Not necessarily the final icon, but better than nothing!

### PR Checklist

### To the reviewers 👀


pvinis (approved):
nice

rajsam003 (approved):
LGTM

rajsam003:
I think we should change the app icon name as well. It should be Folio if I am not wrong. 
image

pvinis:
merge when ready. we can rename it in this PR or in another. should be the PRODUCT_NAME in xcode

rajsam003:
> I renamed to Folio, and then I also enabled iPad because why not? It looks OK so far, and I think we should definitely try and ensure that we make it usable on iPad.

Sure. The artworks grid padding is not proper on iPad. Created [PR](https://github.com/artsy/energy/pull/49) already.

PR #46, opened 2022-05-31
refactor: Separate artists query

rajsam003:
This PR separates the artists fetch query from the Artists component. 

### PR Checklist

### To the reviewers 👀

[docs]: /docs/README.md


src/Screens/HomeTabs/Artists/Artists.tsx
```diff
@@ -10,7 +10,28 @@ import { HomeTabsScreens } from "routes/HomeTabsNavigationStack"
...
+      data={artists}
+      renderItem={({ item: artist }) => (
+        <Touchable
+          onPress={() => {
+            navigation.navigate("ArtistTabs", {
+              artistName: artist.name || "",
+            })
+          }}
+        >
+          <ArtistThumbnail artist={artist as any} />
```
rajsam003, line 26: passing props as any as it was very difficult to type case into `ArtistThumbnailProps`.
Suggestions are appreciated🙏🏽

src/Screens/HomeTabs/Artists/Artists.tsx
```diff
@@ -35,37 +56,18 @@ export const Artists = () => {
...
 interface ArtistThumbnailProps {
-  artist: NonNullable<
-    NonNullable<
-      NonNullable<
-        NonNullable<
-          NonNullable<ArtistsQuery["response"]["partner"]>["allArtistsConnection"]
-        >["edges"]
-      >[0]
-    >["node"]
-  >
```
rajsam003, line 68: This is messy

src/Screens/HomeTabs/Artists/Artists.tsx
MounirDhahri: I think this is because `ArtistThumbnail` is defined as a component instead of a fragment container, I would strongly suggest to rewrite the latter to specify the props it needs

src/Screens/HomeTabs/Artists/Artists.tsx
```diff
@@ -10,7 +10,28 @@ import { HomeTabsScreens } from "routes/HomeTabsNavigationStack"
...
 
+  return (
+    <TabsFlatList
+      data={artists}
+      renderItem={({ item: artist }) => (
+        <Touchable
+          onPress={() => {
+            navigation.navigate("ArtistTabs", {
+              artistName: artist.name || "",
+            })
```
MounirDhahri, line 23: **thought:** shouldn't this navigate to the `Artist` screen instead of `ArtistsTabs` screen. Maybe we need to rename the screen

src/Screens/HomeTabs/Artists/Artists.tsx
MounirDhahri: I believe my comment above should solve this issue and we can get to define this component props as we do for other relay components

src/Screens/HomeTabs/Artists/Artists.tsx
pvinis: we can pair on this tomorrow sam. im sure we can find some way for a nice type.

src/Screens/HomeTabs/Artists/Artists.tsx
```diff
@@ -10,7 +10,28 @@ import { HomeTabsScreens } from "routes/HomeTabsNavigationStack"
...
+        >
+          <ArtistThumbnail artist={artist as any} />
+        </Touchable>
+      )}
+      keyExtractor={(item) => item?.internalID!}
+    />
+  )
+}
+
+const artistsQuery = (partnerID: string) => {
```
pvinis, line 34: this is a bit strange-looking to me. maybe we can look together? my main concern is, what is easier to use:

```
const Artists = () => {
  const partnerID = GlobalStore..
  const artistsData = useLazyLoadQuery(artistsQuery, { partnerID })
  const artists = extractNodes(artistsData.partner?.allArtistsConnection)

  return <FlatList data={artists} .. />
}

const artistsQuery = graphql`
  query ArtistsQuery($partnerID: String!) {
    blabla
  }
`
```

or what you have but a bit more hook-y
```
const Artists = () => {
  const partnerID = GlobalStore..
  const artists = useArtistsQuery(partnerID)

  return <FlatList data={artists} .. />
}

const useArtistsQuery = (partnerID: string) => {
  const data = useLazyLoadQuery(artistsQuery, { partnerID })
  return extractNodes(data.partner?.allArtistsConnection)
}

const artistsQuery = graphql`
  query ArtistsQuery($partnerID: String!) {
    blabla
  }
`
```

pvinis: i kinda like the top option, since the hook in the second one is pretty small. the main thing that id like to have separate is the actual query, since its a big block of text thats not really helping much in the understanding of the component. its more of the declaration of data rather than the building of the component. what do you think about the two options (or any other you can come up with)?

PR #53, opened 2022-06-03
feat: Artwork screen

rajsam003:
This PR resolves [MOPLAT-257]

### Description

https://user-images.githubusercontent.com/17421923/171874696-979a2603-c354-418f-a4da-dccb9c2ad310.mov

### PR Checklist

### To the reviewers 👀

[docs]: /docs/README.md

[MOPLAT-257]: https://artsyproduct.atlassian.net/browse/MOPLAT-257?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Screens/_shared/ArtworkGridItem.tsx
```diff
@@ -1,29 +1,40 @@
...
-        }}
-      />
-      <Text italic variant="xs" color="black60" mt={1}>
-        {artwork.title},{" "}
-        <Text variant="xs" color="black60">
-          {artwork.date}
+    <Touchable
+      onPress={() => {
+        navigation.navigate("Artwork", {
+          slug: artwork.slug || "",
```
pvinis, line 19: not blocking, but I wonder if this is not a good practice. if the slug is null, then passing empty string will just cause the next screen to freak out. I wonder if making the touchable not tappable is better? 🤔

also, if we know the slug is there, we could use a !. otherwise I guess the current screen would probably have issues displaying the artwork.

I also wonder if on MP we should make the slug not nullable, because is it ever null?

src/shared/tests/getEssentialProps.tsx
```diff
@@ -0,0 +1,11 @@
+export const getEssentialProps = (params: {} = {}) =>
```
pvinis, line 1: what is this? i don't see this used anywhere 🤔. also, if it is, maybe we turn this into a mock?

src/Screens/_shared/Artwork/Artwork.tsx
```diff
@@ -1 +1,116 @@
...
+import { HomeTabsScreens } from "routes/HomeTabsNavigationStack"
+import { ArtworkQuery } from "__generated__/ArtworkQuery.graphql"
+
+type ArtworkProps = NativeStackScreenProps<HomeTabsScreens, "Artwork">
+
+export const Artwork: React.FC<ArtworkProps> = ({ route, navigation }) => {
+  const { slug } = route.params
+
+  return (
+    <Suspense
```
pvinis, line 16: not for this PR, but now that we see this pattern everywhere, it might be worth giving some time to it to extract it in a comp and reuse it.

something like ArtsySuspense (or a better name 😅) that has the wraps children in a suspense with the fallback defaulting to the activityindicator 🤔

pvinis (approved):
great work! maybe we can remove the unused mock thing, and then feel free to merge 👍

src/Screens/_shared/ArtworkGridItem.tsx
pvinis: as Damon says here https://artsy.slack.com/archives/C2TQ4PT8R/p1654364617093749?thread_ts=1654334440.295129&channel=C2TQ4PT8R&message_ts=1654364617.093749 slug is already non null. and I see it too in the gql files in this PR. so let's just remove the || "" part.

pvinis (changes requested):
added one more comment for the slug

src/Screens/_shared/Artwork/Artwork.tests.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { Artwork } from "./Artwork"
+
+describe("Artwork", () => {
+  it("renders without throwing an error", async () => {
+    renderWithWrappersTL(<Artwork {...getEssentialProps()} />)
+    await mockEnvironmentPayloadMaybe(mockProps)
+  })
+
+  it("renders the list of artists", async () => {
+    const { queryByText } = renderWithWrappersTL(<Artwork {...getEssentialProps()} />)
```
rajsam003, line 13: here @pvinis

src/shared/tests/getEssentialProps.tsx
rajsam003: Its is used [here](https://github.com/artsy/energy/pull/53/files#r889898645). To get navigation and route as props

src/Screens/_shared/Artwork/Artwork.tsx
rajsam003: Sure. I was thinking about it too.

pvinis (approved):

PR #64, opened 2022-06-15
feat: Persist albums

rajsam003:
This PR resolves [MOPLAT-300]

### Description

To enable creation, editing, and deletion of albums in GlobalStore and to wire up with `Create Album` button to use the storage

### PR Checklist

### To the reviewers 👀

[docs]: /docs/README.md

[MOPLAT-300]: https://artsyproduct.atlassian.net/browse/MOPLAT-300?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/screens/HomeTabs/Albums/CreateAlbum.tsx
```diff
@@ -12,10 +12,41 @@ import {
 import { ShadowSeparator } from "palette/elements/Separator/ShadowSeparator"
 import { useSafeAreaInsets } from "react-native-safe-area-context"
 import { HomeTabsScreens } from "app/routes/HomeTabsNavigationStack"
+import { GlobalStore } from "app/store/GlobalStore"
+import { useFormik } from "formik"
+import * as Yup from "yup"
+import uuid from "react-native-uuid"
+import moment from "moment"
```
MounirDhahri, line 19: **thought:** is there a particular reason why you would like to use `moment` here? maybe you can do the same using `luxon`

src/app/screens/HomeTabs/Albums/CreateAlbum.tsx
```diff
@@ -12,10 +12,41 @@ import {
...
+    useFormik<CreateAlbumValuesSchema>({
+      initialValues: {
+        albumName: "",
+      },
+      initialErrors: {},
+      onSubmit: () => {
+        GlobalStore.actions.albums.addAlbum({
+          id: uuid.v4().toString(),
+          createdAt: moment().format(),
+          title: values.albumName,
```
MounirDhahri, line 43: **suggestion:**
```suggestion
          title: values.albumName.trim(),
```

src/app/screens/HomeTabs/Albums/CreateAlbum.tsx
```diff
@@ -12,10 +12,41 @@ import {
...
+const createAlbumSchema = Yup.object().shape({
+  albumName: Yup.string().required("Album name is required"),
+})
 
 export const CreateAlbum = () => {
   const navigation = useNavigation<NavigationProp<HomeTabsScreens>>()
   const insets = useSafeAreaInsets()
+
+  const { handleSubmit, handleChange, values, errors, validateForm } =
+    useFormik<CreateAlbumValuesSchema>({
```
MounirDhahri, line 34: **thought:** I am not sure you need `useFormik` here, maybe `useFormikContext` instead 👀

src/app/screens/HomeTabs/Albums/CreateAlbum.tsx
```diff
@@ -12,10 +12,41 @@ import {
...
+      initialValues: {
+        albumName: "",
+      },
+      initialErrors: {},
+      onSubmit: () => {
+        GlobalStore.actions.albums.addAlbum({
+          id: uuid.v4().toString(),
+          createdAt: moment().format(),
```
MounirDhahri, line 42: **suggestion:** I would suggest storing the ISO date and then formatting it when needed instead.

src/app/screens/HomeTabs/Albums/CreateAlbum.tsx
```diff
@@ -12,10 +12,41 @@ import {
...
+          title: values.albumName,
+        })
+        navigation.navigate("HomeTabs")
```
MounirDhahri, line 45: **issue:** this will navigate to `HomeTabs` anyway even if `AddAlbum` fails right? maybe we can adjust the action to tell if it failed or not and only navigate back to home when we successfully add an album

src/app/screens/HomeTabs/Albums/CreateAlbum.tsx
```diff
@@ -29,12 +60,14 @@ export const CreateAlbum = () => {
...
-            Album name
-          </Text>
-          <Spacer mt={1} />
           <Flex>
-            <Input />
+            <Input
+              title="Album Name"
+              onChangeText={handleChange("albumName")}
+              onBlur={() => validateForm()}
+              defaultValue={values.albumName}
```
MounirDhahri, line 68: interesting use here - i was expecting value

package.json
```diff
@@ -41,7 +41,7 @@
     "jest-extended": "1.2.1",
     "jest-watch-typeahead": "0.6.1",
     "lodash": "^4.17.21",
-    "moment": "^2.29.2",
+    "moment": "^2.29.3",
```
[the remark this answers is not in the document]
patrinoua: Good catch. Do we have a broadly used preferred alternative? [Luxon](https://moment.github.io/luxon/) [Day.js](https://day.js.org/) [date-fns](https://date-fns.org/) [js-Joda](https://js-joda.github.io/js-joda/) [Date](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Date) object ?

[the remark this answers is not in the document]
pvinis: we have luxon already. lets use that.

src/app/store/Models/AlbumsModel.ts
```diff
@@ -0,0 +1,29 @@
+import { action, Action } from "easy-peasy"
+
+interface Album {
+  id: string
+  title: string
+  artworkIds?: string[]
```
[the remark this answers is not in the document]
MounirDhahri: I don't agree with having `?` here, an empty array makes more sense to me

package.json
```diff
@@ -58,6 +58,7 @@
     "react-native-safe-area-context": "^3.3.2",
     "react-native-screens": "^3.9.0",
     "react-native-svg": "^12.1.1",
+    "react-native-uuid": "^2.0.1",
```
pvinis, line 61: not blocking, but i recently found out about https://github.com/ulid/spec, which is i think the same as uuid v7. if we wanna be the cool kids, we could totally use https://github.com/ulid/javascript. but again, not blocking

src/app/store/Models/AlbumsModel.ts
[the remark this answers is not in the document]
pvinis: we can remove the `?` already. a new album with no artworks will just have `[]` here.

MounirDhahri (changes requested):

package.json
rajsam003: Looks good. 
`ulid` is more popular than `react-native-uuid`
but `react-native-uuid` has only 1 open issue. 
Now I am confused which one to use. 😂

src/app/screens/HomeTabs/Albums/CreateAlbum.tsx
MounirDhahri: yes, something like that would be fine - the `goBack()` doesn't make much sense to me though

package.json
pvinis: go with the one you have added already 👍

src/app/store/Models/AlbumsModel.ts
```diff
@@ -0,0 +1,29 @@
...
+  updateAlbum: Action<this, Album>
+  removeAlbum: Action<this, string>
+}
+
+export const AlbumsModel: AlbumsModel = {
+  albums: [],
+  addAlbum: action((state, album) => {
+    state.albums.push(album)
+  }),
+  updateAlbum: action((state, album) => {
```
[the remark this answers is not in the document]
rajsam003: If we do that then I am worried if the updated artwork is at the top or the middle will come to the bottom of the list. 
And in this case, I don't mind passing the whole album as it would have only array of `ids`.
What do you think? If you still think it's not a good idea, I can change it :)

src/app/screens/HomeTabs/Albums/CreateAlbum.tsx
```diff
@@ -12,10 +12,56 @@ import {
...
+            artworkIds: [],
+            createdAt: DateTime.now().toISO(),
+          })
+          setLoading(true)
+          navigation.dispatch({
+            ...CommonActions.reset({
+              index: 0,
+              routes: [{ name: "HomeTabs" }],
+            }),
+          })
```
rajsam003, line 56: @MounirDhahri As you can see I have used `CommonActions` to reset stack navigator. It works really good but it takes the screen to the first tab of the home screen whereas the Album belongs to the third tab. Do you know how to solve this?

src/app/store/Models/AlbumsModel.ts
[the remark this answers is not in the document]
pvinis: i think we should definitely sort them in the list view, so we shuold not care where they are in the array here. either we sort in the view, or we sort in the model, but we sort somewhere, rather than hope we put it in the right place accidentally.

what about this, we keep it as is, and when we do the artwork screens, we can see how it might work best?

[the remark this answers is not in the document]
rajsam003: Sure. Sounds good

src/app/screens/HomeTabs/Albums/CreateAlbum.tsx
rajsam003: I think I can solve this using forwardRef

PR #73, opened 2022-06-24
feat: header

patrinoua:
Screen Shot 2022-06-24 at 17 06 33

Use the header component from sharedUI

### PR Checklist

### To the reviewers 👀

[docs]: /docs/README.md


rajsam003:
Good work
How about using in [HomeTabs] as well.?(https://github.com/artsy/energy/blob/df3e51940aca9dd598ff19587629a3ae7284a100/src/app/screens/HomeTabs/HomeTabs.tsx#L12)

patrinoua:
> Good work How about using in [HomeTabs] as well.?(
> 
> https://github.com/artsy/energy/blob/df3e51940aca9dd598ff19587629a3ae7284a100/src/app/screens/HomeTabs/HomeTabs.tsx#L12
> 
> )

it's a different design that's why I left it out of this change!

MounirDhahri (approved):

patrinoua:
@rajsam003 as per agreement let's do that the second time we have a header without a back button

PR #92, opened 2022-07-12
feat: Edit Album

rajsam003:
This PR resolves [MOPLAT-304]

1. Updating just title

https://user-images.githubusercontent.com/17421923/178707739-d264a0a7-f4fd-4d64-b95d-41035f9b079c.mov

----------------
2. Adding new artworks to the existing album

https://user-images.githubusercontent.com/17421923/178708473-ba2ec354-2afa-4e24-b678-faac325816b1.mov

----------------
3. Deleting existing artworks from the album

https://user-images.githubusercontent.com/17421923/178708892-4c3702b0-255f-4093-8108-b15e10297546.mov

### PR Checklist

### To the reviewers 👀

[docs]: /docs/README.md

[MOPLAT-304]: https://artsyproduct.atlassian.net/browse/MOPLAT-304?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rajsam003:
> oh god, 10 files diff for changing the title 😱

No. hahaha.
Thats just one video

src/app/screens/HomeTabs/Albums/CreateOrEditAlbum/CreateOrEditAlbum.tsx
```diff
@@ -0,0 +1,185 @@
...
+
+const createAlbumSchema = Yup.object().shape({
+  albumName: Yup.string().required("Album name is required"),
+})
+
+type CreateOrEditAlbumRoute = RouteProp<HomeTabsScreens, "CreateOrEditAlbum">
+
+export const CreateOrEditAlbum = () => {
+  const { mode, albumId } = useRoute<CreateOrEditAlbumRoute>().params || {
+    mode: "create",
```
pvinis, line 35: feels like theres a lot of places we change the code flow based on the mode. why not keep these as separate components/screens? whats the benefit of all in one with many forks?

as a side idea, what if we break these down to smaller pieces, maybe some common ones and some not, so each screen can have some deduplicated stuff?

src/app/store/Models/AlbumsModel.ts
```diff
@@ -1,30 +1,48 @@
 import { action, Action } from "easy-peasy"
-import _ from "lodash"
```
pvinis, line 2: nice

src/app/store/Models/AlbumsModel.ts
```diff
@@ -1,30 +1,48 @@
...
 
 export const AlbumsModel: AlbumsModel = {
   sessionState: {
-    selectedArtworks: [],
+    selectedArtworksForCreateAlbum: [],
+    selectedArtworksForEditAlbum: [],
   },
   albums: [],
+  updateAlbumTitle: action((state, { albumId, title }) => {
+    const index = state.albums.findIndex((x) => x.id === albumId)
```
pvinis, line 43: just a random thought, i wonder if we should have made `albums` an object instead of an array 🤔

src/app/screens/HomeTabs/Albums/CreateOrEditAlbum/CreateOrEditAlbum.tsx
```diff
@@ -0,0 +1,185 @@
...
+              albumIds: [albumId],
+              artworkIdsToRemove: selectedArtworksToRemove,
+            })
+            await GlobalStore.actions.albums.clearAllSelectedArtworksForAnExistingAlbum(albumId)
+          } else {
+            await GlobalStore.actions.albums.addAlbum({
+              id: uuid.v4().toString(),
+              title: values.albumName.trim(),
+              artworkIds: selectedArtworks,
+              createdAt: DateTime.now().toISO(),
```
dimatretyak, line 83: **suggestion**: what do you think about not passing the `id` and `createdAt` fields explicitly for the `addAlbum` action, but automatically assigning/adding these values [here](https://github.com/artsy/energy/blob/23dbf85d99491ce86bb7caa28b4cf28d88f32f81/src/app/store/Models/AlbumsModel.ts#L46-L48)?

For example, like this
```javascript
await GlobalStore.actions.albums.addAlbum({
  title: values.albumName.trim(),
  artworkIds: selectedArtworks,
})

...

addAlbum: action((state, album) => {
  const id = uuid.v4().toString()
  const createdAt = DateTime.now().toISO()
  
  state.albums.push({
    id,
    createdAt,
    ...album,
  })
})
```

src/app/screens/HomeTabs/Albums/CreateOrEditAlbum/CreateOrEditAlbumChooseArtworks.tsx
```diff
@@ -11,19 +11,29 @@ import MasonryList from "@react-native-seoul/masonry-list"
...
-  const selectedArtworks = GlobalStore.useAppState(
-    (state) => state.albums.sessionState.selectedArtworks
-  )
+
+  const selectedArtworks =
+    mode === "edit"
+      ? GlobalStore.useAppState((state) => state.albums.sessionState.selectedArtworksForEditAlbum)
+          .find((album) => album.albumId === albumId)
+          ?.artworks.find((artwork) => artwork.artistSlug === slug)?.artworkIds || []
+      : GlobalStore.useAppState((state) => state.albums.sessionState.selectedArtworksForCreateAlbum)
```
dimatretyak, line 33: **question**: maybe we can put this logic in a separate selector or custom hook and return artworks depending on passed mode? 

**Custom selector**
```javascript
const selectedArtworksInModel = GlobalStore.useAppState(getArtworksByMode(mode, albumId))
```

or 

**Custom hook (if it can be done)**
```javascript
const useArtworksByMode = (mode, albumId) => { ... }
const selectedArtworksInModel = useArtworksByMode(mode, albumId)
 
```

src/app/screens/HomeTabs/Albums/CreateOrEditAlbum/CreateOrEditAlbum.tsx
```diff
@@ -0,0 +1,185 @@
...
+                  name: "HomeTabs",
+                  params: {
+                    tabName: "Albums",
+                  },
+                },
+              ],
+            }),
+          })
+        } catch (error) {
+          setLoading(true)
```
dimatretyak, line 102: **suggestion**: I think here we should set `loading` flag to `false`

dimatretyak: or maybe it's better to use `isSubmitting` from [Formik](https://formik.org/docs/api/formik#issubmitting-boolean)

src/app/screens/HomeTabs/Albums/AlbumArtworks.tsx
```diff
@@ -1,27 +1,46 @@
...
+              <EditIcon fill="black100" width={25} height={25} />
+            </Touchable>
+          </Flex>
+        }
+      />
+      <MasonryList
+        testID="artist-artwork-list"
+        contentContainerStyle={{
+          paddingRight: 20,
+          marginTop: 20,
```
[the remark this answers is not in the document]
rajsam003: Should we?

src/app/screens/HomeTabs/Albums/CreateOrEditAlbum/CreateOrEditAlbum.tsx
```diff
@@ -0,0 +1,185 @@
...
+            })
+            await GlobalStore.actions.albums.addArtworksInAlbums({
+              albumIds: [albumId],
+              artworkIdsToAdd: selectedArtworks,
+            })
+            await GlobalStore.actions.albums.removeArtworksFromAlbums({
+              albumIds: [albumId],
+              artworkIdsToRemove: selectedArtworksToRemove,
+            })
+            await GlobalStore.actions.albums.clearAllSelectedArtworksForAnExistingAlbum(albumId)
```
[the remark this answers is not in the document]
rajsam003: Good idea, I will try that

src/app/screens/HomeTabs/Albums/AlbumArtworks.tsx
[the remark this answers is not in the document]
dimatretyak: to tell the truth, I don't know for sure, and that's why I clarified this point with you 😅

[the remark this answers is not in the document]
dimatretyak: If not, then you can ignore the comment

src/app/screens/HomeTabs/Albums/CreateOrEditAlbum/CreateOrEditAlbum.tsx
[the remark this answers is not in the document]
dimatretyak: Thanks!

src/app/screens/HomeTabs/Albums/AlbumArtworks.tsx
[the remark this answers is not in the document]
admbtlr: :ceiling-cat: Yes, please do! Right now these are magic numbers, which is not great. And it's not crazy to imagine that we might want to increase spacing on tablets, for example

src/app/screens/HomeTabs/Albums/CreateOrEditAlbum/CreateOrEditAlbum.tsx
rajsam003: As discussed, there will be a lot of code duplications if we split. so we are keeping it as it is

src/app/screens/HomeTabs/Albums/AlbumArtworks.tsx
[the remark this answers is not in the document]
rajsam003: In that case, there are a lot of places where we need to change. I will work on it in different PR.

src/app/screens/HomeTabs/Albums/CreateOrEditAlbum/CreateOrEditAlbum.tsx
```diff
@@ -0,0 +1,172 @@
...
+export const CreateOrEditAlbum = () => {
+  const { mode, albumId } = useRoute<CreateOrEditAlbumRoute>().params || {
+    mode: "create",
+    albumId: undefined,
+  }
+  const navigation = useNavigation<NavigationProp<HomeTabsScreens>>()
+  const safeAreaInsets = useSafeAreaInsets()
+  const [selectedArtworksToRemove, setSelectedArtworksToRemove] = useState<string[]>([])
+
+  const selectedArtworksInModel = useArtworksByMode(mode)
```
rajsam003, line 41: @dimatretyak I used custom hook as you advised☝️

src/app/store/Models/AlbumsModel.ts
```diff
@@ -1,56 +1,102 @@
...
+  name: string
   artworkIds: string[]
   createdAt: string
 }
+
+interface Artworks {
+  artistSlug: string
+  artworkIds: string[]
+}
+
```
dimatretyak, line 17: **suggestions** and **thoughts**: I agree with [this comment](https://github.com/artsy/energy/pull/92#discussion_r919926094) from Pavlos. It may be more convenient to store the data in a normalized state (something [like this](https://redux.js.org/tutorials/essentials/part-6-performance-normalization#normalized-state-structure)).

These are optional changes, just suggestions 😅

src/app/store/Models/AlbumsModel.ts
```diff
@@ -1,56 +1,102 @@
...
-  addAlbum: Action<this, Album>
+  addAlbum: Action<this, { name: string; artworkIds: string[] }>
   removeAlbum: Action<this, string>
+  editAlbum: Action<this, { albumId: string; name: string; artworkIds: string[] }>
   addArtworksInAlbums: Action<this, { albumIds: string[]; artworkIdsToAdd: string[] }>
-  removeArtworksFromAlbums: Action<this, { albumIds: string[]; artworkIdsToRemove: string[] }>
-  selectArtworksForANewAlbum: Action<this, this["sessionState"]["selectedArtworks"]>
-  clearAllSelectedArtworks: Action<this>
+  selectArtworksForANewAlbum: Action<this, Artworks>
+  selectArtworksForAnExistingAlbum: Action<this, Artworks | null>
```
dimatretyak, line 29: **nitpick**: maybe use just `selectArtworksForNewAlbum` and `selectArtworksForExistingAlbum` (without **A** and **An**). Feel free to ignore this comment If you do not agree with it

src/app/screens/HomeTabs/Albums/CreateOrEditAlbum/CreateOrEditAlbum.tsx
```diff
@@ -0,0 +1,172 @@
...
+    if (mode === "edit") {
+      GlobalStore.actions.albums.selectArtworksForAnExistingAlbum(null)
+    }
+  }
+
+  return (
+    <Flex flex={1} pt={safeAreaInsets.top}>
+      <Header
+        label={mode === "edit" ? "Edit Album" : "Create Album"}
+        backButtonAction={backButtonAction}
```
dimatretyak, line 116: as far as I understand, this logic will not work if the user returns to the previous page using swipe back gesture or hardware back button on Android

src/app/screens/HomeTabs/Albums/CreateOrEditAlbum/CreateOrEditAlbum.tsx
```diff
@@ -0,0 +1,172 @@
...
+
+  const selectedArtworksInModel = useArtworksByMode(mode)
+  const albums = GlobalStore.useAppState((state) => state.albums.albums)
+  const album = albums.find((album) => album.id === albumId)
+  const selectedArtworks = uniq([...(album?.artworkIds || []), ...(selectedArtworksInModel || [])])
+
+  const { handleSubmit, handleChange, values, errors, validateForm, isValid, dirty, isSubmitting } =
+    useFormik<CreateAlbumValuesSchema>({
+      initialValues: {
+        albumName: mode === "edit" ? album?.name || "" : "",
```
dimatretyak, line 49: **suggestion**: `album?.name ?? ""`

dimatretyak (approved):
Good work👌
Left some comments and suggestions, but feel free to ignore them if you don't agree with them

src/app/store/Models/AlbumsModel.ts
```diff
@@ -1,56 +1,102 @@
...
     state.albums = state.albums.filter((x) => x.id !== albumId)
   }),
-  addArtworksInAlbums: action((state, { albumIds, artworkIdsToAdd }) => {
-    albumIds.forEach((albumId) => {
-      const index = state.albums.findIndex((x) => x.id === albumId)
-      state.albums[index].artworkIds = [...state.albums[index].artworkIds, ...artworkIdsToAdd]
-    })
+  editAlbum: action((state, { albumId, name, artworkIds }) => {
+    const index = state.albums.findIndex((x) => x.id === albumId)
+    if (index !== -1) {
```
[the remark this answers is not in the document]
pvinis: i agree. lets do this @rajsam003, something like:
```
if (index === -1) {
  throw new Error("album not found") // or something
}
.. rest of the code as you have it, with state.ablums[index].name = name etc
```

src/app/store/Models/AlbumsModel.ts
```diff
@@ -1,56 +1,102 @@
...
     })
   }),
-  selectArtworksForANewAlbum: action((state, uncommittedArtworkIds) => {
-    state.sessionState.selectedArtworks.push(...uncommittedArtworkIds)
-    state.sessionState.selectedArtworks = _.uniq(state.sessionState.selectedArtworks)
+  selectArtworksForANewAlbum: action((state, artworks) => {
+    const artistIndex = state.sessionState.selectedArtworksForCreateAlbum.findIndex(
+      (artwork) => artwork.artistSlug === artworks.artistSlug
+    )
+    if (artistIndex !== -1) {
```
pvinis, line 74: to @dimatretyak's point, this code here would be simpler if we turned this array into an object. but maybe we can do that when we touch that code next time, when we merge the add and delete components.

src/app/hooks/useArtworksByMode.tsx
```diff
@@ -0,0 +1,21 @@
...
+    ? GlobalStore.useAppState(
+        (state) =>
+          state.albums.sessionState[type].find((artwork) => artwork.artistSlug === slug)?.artworkIds
+      )
+    : GlobalStore.useAppState((state) => state.albums.sessionState[type]).flatMap(
+        (works) => works.artworkIds
+      )
+}
+
+export const useArtworksByMode = (mode: "create" | "edit", slug?: string) => {
```
pvinis, line 15: i dont like this hook here. this is such a specific hook, and its used only twice, so its basically a code shortcut.

its fine to keep it, but lets move it out of the app level hooks, and maybe inside the hometabs/albums dir, where it is used.

src/app/hooks/useArtworksByMode.tsx
```diff
@@ -0,0 +1,21 @@
...
+const getSelectedArtworksFromAlbum = (type: keyof AlbumsModel["sessionState"], slug?: string) => {
+  return slug
+    ? GlobalStore.useAppState(
+        (state) =>
+          state.albums.sessionState[type].find((artwork) => artwork.artistSlug === slug)?.artworkIds
+      )
+    : GlobalStore.useAppState((state) => state.albums.sessionState[type]).flatMap(
+        (works) => works.artworkIds
+      )
+}
```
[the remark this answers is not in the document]
pvinis: true. or even null specifically. its good to type the return type of our hooks, so we are sure we code what we expect.

src/app/sharedUI/Header.tsx
```diff
@@ -5,16 +5,23 @@ type HeaderProps = {
...
+  backButtonAction,
+}: HeaderProps) => {
   const navigation = useNavigation()
   return (
     <Flex px={2}>
       {!withoutBackButton ? (
         <Flex mt={2}>
           <Touchable
             onPress={() => {
+              backButtonAction?.()
```
pvinis, line 24: nice

src/app/store/Models/AlbumsModel.tests.tsx
```diff
@@ -63,18 +84,129 @@ describe("AlbumsModel", () => {
...
-    GlobalStore.actions.albums.removeArtworksFromAlbums({ albumIds, artworkIdsToRemove })
-    const albumState = __globalStoreTestUtils__?.getCurrentState().albums.albums
+      // Nothing gets updated
+      expect(albumState?.map((album) => album.id)).not.toContain(album.albumId)
+    })
 
-    albumState?.forEach((album) => {
-      expect(album.artworkIds).not.toContain("new-artwork-id-1")
-      expect(album.artworkIds).not.toContain("new-artwork-id-2")
+    it("updates name and new artworks Id", () => {
```
pvinis, line 106:
```suggestion
    it("updates name and new artwork ids", () => {
```

src/app/screens/HomeTabs/Albums/CreateOrEditAlbum/CreateOrEditAlbum.tsx
rajsam003: This is so true 😅 I didn't think about it. 🙆🏽‍♀️

rajsam003: Do you have any suggestions? 
Should I use something like `useFocusEffect()` ?

src/app/screens/HomeTabs/Albums/CreateOrEditAlbum/CreateOrEditAlbum.tsx
pvinis: yep. we should never use `||` in our code when we can use `??` instead.

src/app/screens/HomeTabs/Albums/CreateOrEditAlbum/CreateOrEditAlbum.tsx
pvinis: thats probably a good way to do this. Though I would suggest to do this as a separate PR and ticket, so that we can put this in the Header, and have it hook into the hw back of android, as well as the backswipe.
🤔  what do we think? here is also ok, it just feels separate enough to be done in a new PR.

dimatretyak: unfortunately, `useFocusEffect` will not work correctly

https://user-images.githubusercontent.com/3513494/179227127-d4c59e45-8ee0-4483-a076-e2351997a036.mp4

pvinis (changes requested):
just marking this as req changes, as we discussed in the comments here and on slack

src/app/screens/HomeTabs/Albums/CreateOrEditAlbum/CreateOrEditAlbum.tsx
dimatretyak: > can put this in the Header, and have it hook into the hw back of android, as well as the backswipe.
🤔 what do we think? here is also ok, it just feels separate enough to be done in a new PR

I agree with this 👌

dimatretyak: > Do you have any suggestions

Maybe we can use just `useEffect` 

```javascript
useEffect(() => {
  if (mode === "edit") {
    GlobalStore.actions.albums.selectArtworksForAnExistingAlbum(null)
  }
}, [])
```

dimatretyak: or we just have to store the data in the local store for `CreateOrEditAlbum` in `edit` mode. And every time we render this screen, we will have a local store with an initial state

src/app/hooks/useArtworksByMode.tsx
rajsam003: I agree with the hook usage but I dont prefer moving it to `hometabs/albums` because then it will go inside the `screens` folder that will not follow the project structure pattern.

src/app/screens/HomeTabs/Albums/CreateOrEditAlbum/CreateOrEditAlbum.tsx
rajsam003: Now instead of clearing it with backbuttonAction. I am clearing it when clicking on edit button. What do you think?

src/app/screens/HomeTabs/Albums/AlbumArtworks.tsx
```diff
@@ -20,31 +20,44 @@ export const AlbumArtworks = () => {
...
+          },
+        },
+      ]
+    )
+  }
+
+  const editAlbumHandler = () => {
+    GlobalStore.actions.albums.clearSelectedArtworksForEditAlbum()
+    navigation.navigate("CreateOrEditAlbum", { mode: "edit", albumId })
+  }
```
rajsam003, line 47: Here @pvinis @dimatretyak 
Let me know if its a good workaround.

src/app/screens/HomeTabs/Albums/AlbumArtworks.tsx
```diff
@@ -20,31 +20,44 @@ export const AlbumArtworks = () => {
...
-      {
-        text: "Delete",
-        style: "destructive",
-        onPress: async () => {
-          await GlobalStore.actions.albums.removeAlbum(album.id)
-          navigation.goBack()
+  const deleteAlbumHandler = () => {
+    return Alert.alert(
+      "Are you sure you want to delete this album?",
+      "You cannot undo this action",
```
pvinis, line 26:
```suggestion
      "You cannot undo this action.",
```

PR #126, opened 2022-08-01
feat: Artwork screen layout

rajsam003:
This PR resolves [MOPLAT-333]

This PR adds a drawer to the artwork screen to view the artwork details. 

https://user-images.githubusercontent.com/17421923/182145734-2eb56067-4701-43e1-b0bf-9fbb97f73044.mov

### PR Checklist

### To the reviewers 👀

[docs]: /docs/README.md


src/app/sharedUI/screens/Artwork/ArtworkContent/ArtworkContent.tsx
```diff
@@ -1,102 +1,133 @@
...
-      ) : null}
+                {numberOfAlbumsTheArtworkAvailable === albums.length ? null : (
+                  <ArrowRightIcon ml="auto" fill="black100" />
+                )}
+              </Flex>
+              <Spacer mt={3} />
+            </Touchable>
+          ) : null}
+        </ScrollView>
+      </BottomSheet>
```
rajsam003, line 130: Only changes

src/app/sharedUI/screens/Artwork/ArtworkContent/ArtworkContent.tsx
```diff
@@ -1,102 +1,133 @@
...
+          <ImagePlaceholder height={400} />
+        )}
+      </Flex>
+      <BottomSheet
+        ref={bottomSheetRef}
+        snapPoints={snapPoints}
+        onChange={handleSheetChanges}
+        backgroundStyle={{ borderTopWidth: 1, borderColor: color("black10") }}
+      >
+        <ScrollView style={{ paddingHorizontal: space(2) }} scrollEnabled={isScrollEnabled}>
```
rajsam003, line 66: Changes are only on these lines

src/app/sharedUI/screens/Artwork/Artwork.tsx
```diff
@@ -1,46 +1,30 @@
...
 export const Artwork = () => {
   const { params } = useRoute<ArtworkRoute>()
   const artworkSlugs = params.contextArtworkSlugs ?? [params.slug]
   const insets = useSafeAreaInsets()
-  const navigation = useNavigation<NavigationProp<HomeTabsScreens>>()
-  const space = useSpace()
   const screens: ScrollableScreenEntity[] = artworkSlugs.map((slug) => ({
     name: slug,
     content: (
       <SuspenseWrapper key={slug}>
```
patrinoua, line 18: I wonder if we can remove the SuspenseWrapper here after [this PR](https://github.com/artsy/energy/pull/85)

rajsam003: That's true. It was added by mistake. I will remove it.

patrinoua:
Looks great! I'm just wondering if we should hide the price if it's $0, since it's probably just information missing.

patrinoua (approved):

src/app/sharedUI/screens/Artwork/ArtworkContent/ArtworkContent.tsx
```diff
@@ -1,102 +1,133 @@
...
-}
-
-export const ArtworkContent: React.FC<ArtworkContentProps> = (props) => {
-  const { slug } = props
+export const ArtworkContent = ({ slug }: { slug: string }) => {
   const navigation = useNavigation<NavigationProp<HomeTabsScreens>>()
+  const [isScrollEnabled, setIsScrollEnabled] = useState(false)
+  const color = useColor()
+  const space = useSpace()
+  const bottomSheetRef = useRef<BottomSheet>(null)
```
dimatretyak, line 28: **suggestion**: can we remove it? as far as I can see from the code, this is not used anywhere

src/app/sharedUI/screens/Artwork/ArtworkContent/ArtworkContent.tsx
```diff
@@ -1,102 +1,133 @@
...
   const numberOfAlbumsTheArtworkAvailable = useMemo(() => {
     return albums.filter((album) => album.artworkIds.includes(artworkData.artwork?.internalID!))
       .length
   }, [albums])
 
+  const DRAWER_HEIGHT = 180
+  const SCREEN_HEIGHT = useScreenDimensions().height
+  const IMAGE_FLEX_HEIGHT = (SCREEN_HEIGHT - DRAWER_HEIGHT) / SCREEN_HEIGHT
+
+  const snapPoints = useMemo(() => [DRAWER_HEIGHT, SCREEN_HEIGHT], [])
```
dimatretyak, line 41: **suggestion**: maybe we need to add `SCREEN_HEIGHT` in deps (see attached video).

https://user-images.githubusercontent.com/3513494/182187141-8b654d1a-e677-4004-b195-689bfcaa23f3.mp4

Something like this
```javascript
const snapPoints = useMemo(() => [DRAWER_HEIGHT, SCREEN_HEIGHT], [SCREEN_HEIGHT])
```

src/app/sharedUI/screens/Artwork/ArtworkContent/ArtworkContent.tsx
```diff
@@ -1,102 +1,133 @@
...
+          <Spacer mt={0.5} />
+          <Text weight="medium">{artworkData.artwork?.price || "$0"}</Text>
+          <Spacer mt={0.5} />
           <Text variant="xs" color="black60">
-            {artworkData.artwork?.inventoryId}
+            {artworkData.artwork?.mediumType?.name}
           </Text>
+          {(artworkData.artwork?.dimensions?.in || artworkData.artwork?.dimensions?.cm) && (
+            <Text variant="xs" color="black60">
+              {artworkData.artwork?.dimensions?.in} - {artworkData.artwork?.dimensions?.cm}
```
dimatretyak, line 79: **question**: maybe some placeholder value should be rendered if there is no value? Because now I see a situation where, for example, there is no value of `cm` and as a result we get the following `70 9/10 × 70 9/10 in -`

src/app/sharedUI/screens/Artwork/Artwork.tsx
```diff
@@ -1,46 +1,25 @@
...
   const screens: ScrollableScreenEntity[] = artworkSlugs.map((slug) => ({
     name: slug,
-    content: (
-      <SuspenseWrapper key={slug}>
-        <ScrollView style={{ paddingHorizontal: space(2) }}>
-          <ArtworkContent slug={slug} />
-        </ScrollView>
-      </SuspenseWrapper>
-    ),
+    content: <ArtworkContent slug={slug} />,
```
dimatretyak, line 16: If we don't wrap each screen in a `SuspenseWrapper`, then the already fetched and rendered content will be overlapped by the loading indicator. But maybe I don't understand something and thanks to [these changes](https://github.com/artsy/energy/pull/85) we don't have to wrap every screen

https://user-images.githubusercontent.com/3513494/182187842-2f382237-dfa9-40e0-ba86-41b46a6818c9.mp4

src/app/sharedUI/screens/Artwork/ArtworkContent/ArtworkContent.tsx
```diff
@@ -1,102 +1,133 @@
...
+          <Image
+            source={{ uri: artworkData.artwork?.image?.url }}
+            style={{ flex: 1 }}
+            resizeMode="contain"
+          />
+        ) : (
+          <ImagePlaceholder height={400} />
+        )}
+      </Flex>
+      <BottomSheet
```
dimatretyak, line 60: **question**: maybe we shouldn't render the `bottom sheet` for each screen. Maybe it would be better to render this component only once (e.g. in parent container) and display the corresponding content for the active artwork.

src/app/sharedUI/screens/Artwork/ArtworkContent/ArtworkContent.tsx
dimatretyak: or maybe it's better to do something like that?

```javascript
const getDimensionsLabel = () => {
  const inches = artworkData.artwork?.dimensions?.in
  const centimeters = artworkData.artwork?.dimensions?.cm

  if (inches && centimeters) {
    return `${inches} - ${centimeters}`
  }

  if (inches) {
    return inches
  }

  if (centimeters) {
    return centimeters
  }

  return null
}

const dimensions = getDimensionsLabel()

...

{dimensions && (
  <Text variant="xs" color="black60">
    {dimensions}
  </Text>
)}
```

src/app/sharedUI/screens/Artwork/Artwork.tsx
rajsam003: Ahh, this was the reason. I thought it was added by mistake. Let me revert this line.

src/app/sharedUI/screens/Artwork/ArtworkContent/ArtworkContent.tsx
rajsam003: Ya good idea. I will try this too

src/app/sharedUI/screens/Artwork/ArtworkContent/ArtworkContent.tsx
rajsam003: Good catch.
This change is not in the scope of this PR/ticket. Can I do it in a different PR? What do you think?

dimatretyak: yep, sure 😅

src/app/sharedUI/screens/Artwork/ArtworkContent/ArtworkContent.tsx
rajsam003: As discussed, we are using this approach in a different PR

dimatretyak (approved):
🚀🚀🚀

src/app/sharedUI/screens/Artwork/ArtworkContent/ArtworkContent.tests.tsx
```diff
@@ -6,6 +6,15 @@ jest.mock("@react-navigation/native", () => ({
   useNavigation: jest.fn(),
 }))
 
+jest.mock("shared/hooks", () => {
```
pvinis, line 9: this might be better if it lived in the setupJest file rather than here

src/app/sharedUI/screens/Artwork/ArtworkContent/ArtworkContent.tsx
```diff
@@ -1,102 +1,131 @@
...
+
   const albums = GlobalStore.useAppState((state) => state.albums.albums)
   const artworkData = useLazyLoadQuery<ArtworkContentQuery>(artworkContentQuery, { slug })
   const numberOfAlbumsTheArtworkAvailable = useMemo(() => {
     return albums.filter((album) => album.artworkIds.includes(artworkData.artwork?.internalID!))
       .length
   }, [albums])
 
+  const DRAWER_HEIGHT = 180
+  const SCREEN_HEIGHT = useScreenDimensions().height
```
pvinis, line 37: why not
```suggestion
  const SCREEN_HEIGHT = "100%"
```
?

src/app/sharedUI/screens/Artwork/ArtworkContent/ArtworkContent.tsx
```diff
@@ -1,102 +1,131 @@
...
+  const IMAGE_FLEX_HEIGHT = (SCREEN_HEIGHT - DRAWER_HEIGHT) / SCREEN_HEIGHT
```
pvinis, line 38: what does this do? i dont understand this 🤔

src/app/sharedUI/screens/Artwork/ArtworkContent/ArtworkContent.tsx
```diff
@@ -1,102 +1,131 @@
...
+
+  const snapPoints = useMemo(() => [DRAWER_HEIGHT, SCREEN_HEIGHT], [SCREEN_HEIGHT])
+
+  const handleSheetChanges = useCallback((index: number) => {
+    setIsScrollEnabled(!!index)
```
pvinis, line 43: so it this basically enabling scrolling if the snap point is not the bottom snap point? 🤔

pvinis (changes requested):
nice! a couple of questions and some tiny changes, but nice. also from slack, check if we can make the transition from bottom snap point to full screen by tapping.

src/app/sharedUI/screens/Artwork/ArtworkContent/ArtworkContent.tsx
rajsam003: This calculates the flex value for the image based on screen height

src/app/sharedUI/screens/Artwork/ArtworkContent/ArtworkContent.tsx
rajsam003: Yes. In other words, enable scroll only when the bottom sheet gets expanded.

src/app/sharedUI/screens/Artwork/ArtworkContent/ArtworkContent.tsx
rajsam003: It's also easier to calculate the flex value for the image container.

src/app/sharedUI/screens/Artwork/ArtworkContent/ArtworkContent.tests.tsx
rajsam003: Yes, I put it there first, but then other tests started failing for some reason. 😅

pvinis: i would say lets see these fails and check if we can make them work 🤔

src/app/sharedUI/screens/Artwork/ArtworkContent/ArtworkContent.tsx
pvinis: lets put a comment on this, so its clear?

pvinis (approved):

PR #127, opened 2022-08-01
feat(MOPLAT-335): Show artwork in CMS

patrinoua:
### Description

This PR resolves [MOPLAT-335]

Follow up: add the right link, updates to metaphysics are required.

### PR Checklist

- [ ] I tested my changes on **iOS** and **Android**.
- [ ] I added screenshots or videos to illustrate my changes.
- [ ] I added Tests and Stories for my changes.

### To the reviewers 👀

- [ ] I would like **at least one** of the reviewers to run this PR on the simulator or device.

Need help with something? Have a look at our [docs], or get in touch with us.

[docs]: /docs/README.md

[MOPLAT-335]: https://artsyproduct.atlassian.net/browse/MOPLAT-335?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #130, opened 2022-08-08
feat(MOPLAT-335): Artwork: Open artwork in CMS - step by step

patrinoua:
### Description

This PR resolves [MOPLAT-335]
replaces https://github.com/artsy/energy/pull/127

Follow up: add the right link, updates to metaphysics are required.
### PR Checklist
### To the reviewers 👀

[docs]: /docs/README.md

[MOPLAT-335]: https://artsyproduct.atlassian.net/browse/MOPLAT-335?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


patrinoua:
@pvinis @rajsam003 thank you for reviewing https://github.com/artsy/energy/pull/127. 

Since there were quite some questions/suggestions I went ahead and implemented this from scratch so I can provide some more context step-to-step on why some things were added.

When using WebView we need to link it and run pod install. [here](https://www.educative.io/answers/how-to-use-webview-in-react-native) is some more info about that, although they don't go much into details.

Otherwise we are getting this error
Screen Shot 2022-08-08 at 13 08 32

Some info about this also here https://stackoverflow.com/questions/61832232/invariant-violation-requirenativecomponent-rncwebview-was-not-found-in-the-u

src/app/sharedUI/screens/Artwork/ArtworkContent/ArtworkContent.tsx
```diff
@@ -144,6 +146,19 @@ export const ArtworkContent = ({ slug }: { slug: string }) => {
...
           ) : null}
+          <Spacer mt={4} />
+          <Button
+            block
+            onPress={() =>
+              navigation.navigate("ArtworkWebView", {
+                uri: `https://cms-staging.artsy.net/artworks/${artworkData.artwork?.internalID}/edit`,
+              })
+            }
+            icon={<EditIcon fill="white100" width="25" height="25" />}
```
patrinoua, line 157: @pvinis I would love us to have different size variations for our icons, eg "small" "medium" "large" as well as a default size (eg medium) so that we don't always pass these numbers hardcoded. What do you think?

patrinoua: We could have a wrapper <Icon> component that does that

rajsam003 (approved):
LGTM

ios/Podfile
```diff
@@ -21,6 +21,8 @@ target 'Energy' do
     :hermes_enabled => false
   )
 
+  pod 'react-native-webview', :path => '../node_modules/react-native-webview'
```
pvinis, line 24: can you try pod installing but without this line? i really think this is not needed here.

pvinis (changes requested):

src/app/sharedUI/screens/Artwork/ArtworkContent/ArtworkContent.tsx
pvinis: i would first ask our designers, as this is totally their decision. If we only need small or large in a couple of places, maybe its not worth making all icons have three scales, and we can just do what you did here.

src/app/sharedUI/screens/Artwork/ArtworkContent/ArtworkContent.tsx
```diff
@@ -144,6 +146,19 @@ export const ArtworkContent = ({ slug }: { slug: string }) => {
               <Spacer mt={3} />
             </Touchable>
           ) : null}
+          <Spacer mt={4} />
+          <Button
+            block
+            onPress={() =>
+              navigation.navigate("ArtworkWebView", {
+                uri: `https://cms-staging.artsy.net/artworks/${artworkData.artwork?.internalID}/edit`,
```
pvinis, line 154: this works now, without the partnerid param?

ios/Podfile
patrinoua: This was automatically added when running the commands, I did not add it manually. I can manually remove it but then please test this PR locally to make sure it runs properly without this.

src/app/sharedUI/screens/Artwork/ArtworkContent/ArtworkContent.tsx
patrinoua: added 🙌🏻

brainbicycle:
I pushed a couple small changes:
- I added a header with a back button, without this can't get back to the app from the webview
- I changed the url to use slug instead of internal id, this works after authenticating the first time

One minor follow-up:
- We should probably steal the loading indicator from Eigen's webview, sorry this was overlooked in the ticket but it just shows a white screen while loading, I'll create a ticket for that

patrinoua:
@brainbicycle thanks!!

pvinis:
before merging, let's check if we can remove the podfile and android changes.

patrinoua:
replaced by https://github.com/artsy/energy/pull/133

PR #146, opened 2022-08-25
perf: Replacing artwork(id) with artworks query

rajsam003:
This PR improves the performance of the app by fetching all artworks at once instead of nesting query renderers for the albums. 

Before: 
```
artwork(id: $id) {
      ...ArtworkGridItem_artwork
    }

```

After:
```
artworks(ids: [id1, id2, id3, ...]) {
      edges{
           node{
          ...ArtworkGridItem_artwork
          }
       }
    }
```

### PR Checklist

### To the reviewers 👀

[docs]: /docs/README.md


src/app/screens/HomeTabs/Albums/CreateOrEditAlbum/CreateOrEditAlbum.tsx
```diff
@@ -44,6 +48,18 @@ export const CreateOrEditAlbum = () => {
...
+    To avoid that we are giving a condition to only call the API when the selectedArtworks is not empty*/
+  const artworks =
+    selectedArtworks.length > 0
+      ? extractNodes(
+          useLazyLoadQuery<AlbumArtworksQuery>(albumArtworksQuery, {
+            artworkIds: selectedArtworks,
+          }).artworks
+        )
+      : []
+
```
rajsam003, line 62: @mdole Here is the case I was talking about☝️

src/app/screens/HomeTabs/Albums/CreateOrEditAlbum/CreateOrEditAlbum.tsx
```diff
@@ -44,6 +48,18 @@ export const CreateOrEditAlbum = () => {
   const selectedArtworks = uniq([...(album?.artworkIds || []), ...selectedArtworksInModel])
   const space = useSpace()
 
+  /* If we pass empty ids(selectedArtworks) as parameter in the albumArtworksQuery API,
```
MounirDhahri, line 51: **praise:** thanks for the comment, it makes it easier to understand your reasoning here.

MounirDhahri: **suggestion:** I am wondering though if a hotfix is what we need here, I would suggest here to add a new argument to the interface (`allowEmptyReturns` or something like that) to let it know to return nothing if nothing is provided. This way we can fix the problem everywhere in Artsy and make the next dev life a bit easier

Speaking of making a change here https://www.notion.so/artsy/Sprint-17-cdd05a12330e4b27b9c2ba16a8a9d812

MounirDhahri (commented):
Nice PR Sam, thanks for making this change ❤️ I left a small comment, let me know if you didn't get what I mean

src/app/screens/HomeTabs/Albums/CreateOrEditAlbum/CreateOrEditAlbum.tsx
```diff
@@ -44,6 +48,18 @@ export const CreateOrEditAlbum = () => {
...
+    it returns random artworks(instead of returning empty object),
```
patrinoua, line 52: Wow this sounds bad! Thoughts @brainbicycle @mc-jones @lidimayra ?

rajsam003: We already have a thread for this
https://artsy.slack.com/archives/CP9P4KR35/p1661419259015689

brainbicycle: hmmm, it sounds like weird behavior by the api to me but also if we don't have artwork ids I think it makes sense not to query the api at all? so the frontend guard against that condition seems like enough to me

brainbicycle: but I actually think the bigger concern from that thread is the endpoint is deprecated so we may not want to use this at all in folio?

brainbicycle (changes requested):
Thanks Sam for thinking about the performance here! I think unfortunately we may want to keep it as is for now because from the thread it sounds like this endpoint is deprecated. It is possible we could use artworksConnection or filterArworksConnection for the same purpose but I would suggest holding off until we have a better sense of what offline access looks like. Reason being if offline ends up requiring fetching all data up front that might dictate which endpoints we use and also would mean that once fetched going to a page would be instant.

rajsam003:
> Thanks Sam for thinking about the performance here! I think unfortunately we may want to keep it as is for now because from the thread it sounds like this endpoint is deprecated. It is possible we could use artworksConnection or filterArworksConnection for the same purpose but I would suggest holding off until we have a better sense of what offline access looks like. Reason being if offline ends up requiring fetching all data up front that might dictate which endpoints we use and also would mean that once fetched going to a page would be instant.

Makes total sense Brian

rajsam003:
> Should we close this pr?

Ya sure. I am closing it.

PR #157, opened 2022-08-30
feat(MOPLAT-262): Install Shot QA

patrinoua:
### Description

This PR resolves [MOPLAT-262]

- adds positionAbsolute to header so that it is not calculated relative to the position of other elements on the screen

center install image 
Screen Shot 2022-08-30 at 11 43 06

### PR Checklist

### To the reviewers 👀

[docs]: /docs/README.md

[MOPLAT-262]: https://artsyproduct.atlassian.net/browse/MOPLAT-262?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/sharedUI/screens/InstallImage/InstallImage.tsx
```diff
@@ -12,7 +12,7 @@ export const InstallImage = () => {
 
   return (
     <Flex flex={1} mt={2}>
-      <Header safeAreaInsets />
+      <Header safeAreaInsets positionAbsolute />
```
rajsam003, line 15: Header is getting more props 😁, which I dont think is a good idea. Because we'll soon be working on removing these extra props from Header.
How about adding `position:absolute` to the Flex below it?

patrinoua: How about we leave it as it is for now and address a different approach when we work on the Header?

patrinoua: I am suggesting this because I think we should avoid using "absolute" position, generally as much as possible.
Adding it as a prop to the header seems more reasonable, instead of the main screen.
But we can definitely think about the Header more once we refactor it.

patrinoua: It also does seem like position absolute on the header could be used in other places as well.

rajsam003: Yes sounds good. We can discuss and refactor this as well while working on the Header fix ticket.

brainbicycle (approved):

PR #165, opened 2022-08-31
feat(MOPLAT-352): Util function to retrieve the right size for artworks

patrinoua:
### Description

This PR resolves [MOPLAT-352]

Attention: It's way slower when loading artworks. 

### PR Checklist
### To the reviewers 👀

[docs]: /docs/README.md

[MOPLAT-352]: https://artsyproduct.atlassian.net/browse/MOPLAT-352?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/__generated__/ArtistArtworksQuery.graphql.ts
```diff
@@ -53,7 +53,9 @@ fragment ArtworkGridItem_artwork on Artwork {
   title
   date
   image {
-    url
+    resized(version: "normalized") {
```
rajsam003, line 56: **non-blocking** - Is .gitignore not working for you?

src/app/utils/imageSize.ts
```diff
@@ -0,0 +1,6 @@
+import { Dimensions, PixelRatio } from "react-native"
+
+const windowWidth = Dimensions.get("window").width
+const pixelRatio = PixelRatio.get()
+
+export const imageSize = 2 * windowWidth * pixelRatio
```
rajsam003, line 6: **thought**: This calculation has nothing to do with image so I dont think `imageSize` is a good name. May be `screenWidth` or `screenSize`.

src/app/sharedUI/screens/Artwork/ArtworkContent/ArtworkContent.tsx
```diff
@@ -303,10 +307,12 @@ const ArtworkDetail = ({ size = "small", label, value }: ArtworkDetailProps) =>
 }
 
 const artworkContentQuery = graphql`
-  query ArtworkContentQuery($slug: String!) {
+  query ArtworkContentQuery($slug: String!, $imageSize: Int!) {
     artwork(id: $slug) {
       image {
-        url
+        resized(height: $imageSize, version: "normalized") {
```
rajsam003, line 313: **question**: We are calculating for `width` and passing it for height. Did you try passing it for width. I mean - `resized(width: $imageSize, version: "normalized")`

src/__generated__/ArtistArtworksQuery.graphql.ts
patrinoua: probably not! Strange right?

src/app/utils/imageSize.ts
patrinoua: It is the imageSize we wish for. It uses screen width, but it is not screen width or screen size.  Or if you think it is better we can rename it to "calculatedImageWidthForDoublePersision" but that could be a little bit of an overkill, so we may as well just keep "imageSize" as a convention.

Open to suggestions anyone has a better name idea! 🤷🏻‍♀️

src/app/sharedUI/screens/Artwork/ArtworkContent/ArtworkContent.tsx
patrinoua: Good point, I've updated that to use the window height instead

 
Screen Shot 2022-09-01 at 14 24 43

src/__generated__/ArtistArtworksQuery.graphql.ts
rajsam003: Yes👀. For me, it's working fine.

rajsam003 (approved):
Good work and thanks for addressing my comments :) 
🚀

src/app/sharedUI/screens/Artwork/ArtworkContent/ArtworkContent.tsx
admbtlr: Pretty sure we want to use windowWidth here, not windowHeight

admbtlr: It's line 313 that should change here: 
```suggestion
        resized(width: $imageSize, version: "normalized") {
```

src/app/utils/imageSize.ts
```diff
@@ -0,0 +1,6 @@
+import { Dimensions, PixelRatio } from "react-native"
+
+const windowHeight = Dimensions.get("window").height
+const pixelRatio = PixelRatio.get()
+
+export const imageSize = 2 * windowHeight * pixelRatio
```
admbtlr, line 6: And then this should go back to using `windowWidth` instead of `windowHeight`

admbtlr (changes requested):

src/app/utils/imageSize.ts
patrinoua: I am happy to update this, however I am curious, since we are not checking if width or height is bigger and this may differ from one image to the other, why do we prefer width over height? 
It felt like an axiomatic choice but curious to see if there is a reasoning behind this.

patrinoua: Thanks. I've updated to use width, feel free to merge if it's ok.

src/app/utils/imageSize.ts
pvinis: non blocking, but i would prefer `calculatedImageWidthForDoublePersision` to be honest, because `imageSize` is a name that could appear naturally in our code, and i wouldn't want my editor to start importing stuff for me when i don't need them.

pvinis (approved):

pvinis:
merge when ready 👍. i assigned it back to you. thanks!

src/app/utils/imageSize.ts
admbtlr: good point - but I guess you mean `calculatedImageWidthForDoublePrecision`?

PR #190, opened 2022-09-09
fix: data inconsistencies

brainbicycle:
### Description

This PR resolves [MOPLAT-371]
This PR resolves [MOPLAT-370]

### PR Checklist

### To the reviewers 👀

[docs]: /docs/README.md


src/app/screens/HomeTabs/Artists/ArtistTabs/ArtistTabs.tsx
```diff
@@ -4,6 +4,7 @@ import { graphql, useLazyLoadQuery } from "react-relay"
 import { ArtistTabsQuery } from "__generated__/ArtistTabsQuery.graphql"
 import { HomeTabsScreens } from "app/navigation/HomeTabsNavigationStack"
 import { Header } from "app/sharedUI"
+import { GlobalStore } from "app/store/GlobalStore"
```
brainbicycle, line 7:
```suggestion
```

src/app/screens/HomeTabs/Artists/Artists.tsx
```diff
@@ -13,33 +15,51 @@ export const Artists = () => {
...
-            })
-          }
-        >
-          <ArtistListItem artist={artist} />
-        </Touchable>
-      )}
+      data={items}
+      renderItem={({ item }) => {
+        const artist = item[0]!
+        const count = item[1]!
```
brainbicycle, line 32: don't love the `!`, anybody have a preferred way of unwrapping safely?

src/app/screens/HomeTabs/Artists/Artists.tsx
```diff
@@ -13,33 +15,51 @@ export const Artists = () => {
   const partnerID = GlobalStore.useAppState((state) => state.activePartnerID)!
   const artistsData = useLazyLoadQuery<ArtistsQuery>(artistsQuery, { partnerID })
   const artists = extractNodes(artistsData.partner?.allArtistsConnection)
+  const counts = artistsData.partner?.allArtistsConnection?.edges?.map(
+    (edge) => edge?.counts?.managedArtworks as string
+  )
+  if (!counts || !artists) {
+    return
```
brainbicycle, line 22: is there a pattern here, should I return an empty component or maybe show an error when we have an error page?

pvinis: we dont have anything yet. we could copy something over from eigen, or just keep it as you have it and we work on this in the future.

src/app/screens/HomeTabs/Artists/Artists.tsx
pvinis: i guess these are optional on gql, so one easy way would be just to return null if they are empty.

an even better way would be something like:
```
data={items.filter(([artist, count]) => count > 0 && !!artist)} // or whatever is the right check here
renderItem={({ item })} => (
  <Touchable ..
```
so that we only render items that have all the things we need to render them.

otherwise `!` is ok for v1 🤷

pvinis (approved):

patrinoua:
The numbers seem right, but the images are not the same. Should we be hitting a different endpoint? 
In this case it seems like the mobile actually works better than the web version. 

Screen Shot 2022-09-13 at 16 54 01

Screen Shot 2022-09-13 at 16 51 10

brainbicycle:
> The numbers seem right, but the images are not the same. Should we be hitting a different endpoint? In this case it seems like the mobile actually works better than the web version, so we should probably update the URL link in volt.
> 
> Screen Shot 2022-09-13 at 16 54 01
> 
> Screen Shot 2022-09-13 at 16 51 10

hmm, good find. interesting not sure the logic cms is using for these images, I'd say old folio and not cms should be the source of truth here but I think those images are different as well. either way I would say we can address that with a follow up pr rather than holding this one up.

patrinoua (approved):

PR #201, opened 2022-09-14
feat: better icons

pvinis:
### Description

This PR resolves []

icons! ✨ 

the main difference here is that the icons with the extra padding around our logo are the ones that appear on the home screen. the ones with smaller padding are the ones that appear on notifications and settings.

Screenshot 2022-09-14 at 17 58 37
Screenshot 2022-09-14 at 17 53 00

### PR Checklist

### To the reviewers 👀

[docs]: /docs/README.md


patrinoua (approved):
Cool! Just a small suggestion, in order to have this information also when we need to work with the icons I would suggest naming them something like
ArtsyIconNotifications-SizeETC
ArtsyIconHomeScreen-SizeETC

patrinoua:
ArtsyIcon would make it easier to find as well.
Also, it would be great to have storybook to display all the variations and add comments for usage :)

patrinoua (changes requested):
Let's keep the name as ArtsyIcon instead of Icon

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  ❌ **It seems like you added some non WebP images to Eigen, please convert them to WebPs using `source images/script.sh` script **
  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 1f5d69a20f4907a969729c0283d8fcfd8a5a6d34
</p>

pvinis:
renamed, ready to merge.

storybook is interesting, we could add that, but not right now. naming, its more coming from apple, not us, and we already need that info when designing, so the filenames are not the place the designers would look at.

patrinoua:
@pvinis sorry for holding this but would you mind calling them ArtsyIcons as they originally were? Thanks 🌼

pvinis:
`AppIcon` is more descriptive than `ArtsyIcon`. Thats what Xcode calls them as well.

patrinoua:
But the App in this case is Artsy (or actually we use the Artsy app for folio, maybe later we have a separate folio icon, who knows). Why do you prefer App?
It is also called Artsy Icon in our design System and palette web and mobile, so for consistency I would insist on calling it ArtsyIcon 🙏 
https://www.figma.com/file/gZNkyqLT8AU3T61tluVJyB/Artsy-3.1-Design-System?node-id=6726%3A1164
https://palette.artsy.net/atoms/icons/
https://github.com/artsy/eigen/tree/main/src/palette/svgs

pvinis:
This is the `AppIcon`, the *icon* of this *app*. Yes, it is the *artsy logo*, but its not the artsy icon. The app icon is the artsy logo in black, on a white background.
The eigen app icon, is also the artsy logo, in white, on a black background.

These are both the app icons, to be used on the ios homescreen. they have nothing to do with the svg icons, or the design system. They are what xcode needs to tell ios to display, and that is an app icon.

To be honest, names dont even matter here, they are just here for xcode to know where to find those files, but if we are getting hung up on names, app icon is what these files are, they dont have any meaning outside xcode, and their content is not what the name should be based on.

Hopefully that makes sense?

patrinoua:
To my understanding we can pass anything to 
ios/Energy/Images.xcassets/AppIcon.appiconset/Contents.json
and it will display it.
Before (on current main branch) we were passing **ArtsyIcon**, while with this PR we are passing **AppIcon**.
So it only has to do with how we refer to things and it's a good practice in general to avoid using different names for the same thing.

brainbicycle (approved):

brainbicycle:
I tend to agree with Pavlos, we have a few icons in our app that could be called Artsy icons, this is the icon for the app, maybe a more accurate name would be ArtsyAppIcon or ArtsyFolioAppIcon 🙃, but frankly I think this doesn't make a big difference either way, for finding this my first thought would be AppIcon.

patrinoua:
This is causing a miss-alignment between how we refer to the different things that exist in our design system/energy/eigen and is not a good practice in general, so I would ask that we try to avoid it in the future, by sticking to naming things after what they are named on our design system.

admbtlr:
But these icons are not part of the design system per se, they're app icons. The design system refers to elements _within_ the app, while these icons exist "outside" the app. `AppIcon` is a standard name, and I think it's better.

patrinoua:
The logo / ArtsyIcon or AppIcon if that's what we want to call it is part of the Design system and can be found in the figma file, palette, eigen etc, and is always called "ArtsyIcon" everywhere except this PR.

[figma.com/file/gZNkyqLT8AU3T61tluVJyB/Artsy-3.1-Design-System?node-id=6726%3A1164](https://www.figma.com/file/gZNkyqLT8AU3T61tluVJyB/Artsy-3.1-Design-System?node-id=6726%3A1164)
[palette.artsy.net/atoms/icons](https://palette.artsy.net/atoms/icons/)
[artsy/eigen@main/src/palette/svgs](https://github.com/artsy/eigen/tree/main/src/palette/svgs?rgh-link-date=2022-09-16T10%3A35%3A51Z)

Screen Shot 2022-09-20 at 11 49 38

PR #214, opened 2022-09-20
fix: bigger tap area on the more icon

pvinis:
### Description

This PR resolves []

this one
SCR-20220920-frf-2

### PR Checklist

### To the reviewers 👀

[docs]: /docs/README.md


src/app/sharedUI/screens/Artwork/Artwork.tsx
```diff
@@ -65,7 +65,10 @@ export const Artwork = () => {
       <Header
         safeAreaInsets
         rightElements={
-          <Touchable onPress={handlePresentModalPress}>
+          <Touchable
+            onPress={handlePresentModalPress}
+            hitSlop={{ top: 10, left: 10, right: 10, bottom: 10 }}
```
patrinoua, line 70: @artsy/mobile-platform Touchable comes from palette so I wonder if instead of passing hitslop everytime we can just update it on the palette component

patrinoua: Or add a "space" boolean property or something like that

PR #234, opened 2022-09-29
Adjust snap points

patrinoua:
### Before

Screen Shot 2022-09-29 at 17 27 58

### After
Screen Shot 2022-09-29 at 17 27 00

### PR Checklist

### To the reviewers 👀

[docs]: /docs/README.md


src/app/sharedUI/screens/Artwork/Artwork.tsx
```diff
@@ -48,7 +48,7 @@ export const Artwork = () => {
 
   const height = Dimensions.get("window").height
 
-  const snapPoints = useMemo(() => [height > 800 ? "65%" : "75%"], [])
+  const snapPoints = useMemo(() => [height > 1000 ? "45%" : height > 800 ? "65%" : "75%"], [])
```
rajsam003, line 51: I dont think this is an ideal solution. Let's pair on this if you want :)

rajsam003, line 51:
```suggestion
  const snapPoints = useMemo(() => [500], [])
```
This will make it look nicer on all devices. We dont need height.

brainbicycle: if 500 is enough for both device types sure but either way seems okay to me if it is better to be dynamic based on. device size

brainbicycle (approved):
maybe can simplify according to Sam's comment if it is enough for both devices but looks good to me either way, thanks!

patrinoua:
I think the dynamic sizing looks better but don't mind going for any of the two

## 500
Screen Shot 2022-09-30 at 16 01 22

## dynamic

Screen Shot 2022-09-30 at 16 01 52

PR #241, opened 2022-09-30
feat: faster login if user is a dev 🙌🏻

patrinoua:
### Description

Prefill email and password fields if user is a developer

### PR Checklist

### To the reviewers 👀

[docs]: /docs/README.md


src/app/screens/Auth/Login.tsx
```diff
@@ -199,7 +209,7 @@ export const LoginScreen = () => {
               />
             </>
           )}
-          {!showOtpInputField && (
+          {/* {!showOtpInputField && (
```
brainbicycle, line 212: this still meant to be here?

src/app/screens/Auth/Login.tsx
```diff
@@ -149,6 +155,7 @@ export const LoginScreen = () => {
             autoCorrect={false}
             onChangeText={(text) => {
               // Hide error when the user starts to type again
+              GlobalStore.actions.auth.saveDevPassword(text)
```
brainbicycle, line 158: maybe better to save on submit then onChange?

patrinoua: I just tried but it seems like this would make the logic required to check between dev or not dev more complicated

src/app/screens/Auth/Login.tsx
patrinoua: Yes! This functionality is not implemented yet so I thought we can comment it out until we have it

src/app/screens/Auth/Login.tsx
brainbicycle: alright cool it should be fine this way 👍

brainbicycle (approved):

PR #242, opened 2022-09-30
feat: MOPLAT-417 Send Email pt 2: Send Artwork(s) via email

patrinoua:
### Description

This PR resolves [MOPLAT-417]

### PR Checklist
### To the reviewers 👀

[docs]: /docs/README.md

https://user-images.githubusercontent.com/36475005/193308189-b14f0d83-1ccd-4ead-9aa3-311b833dcd13.MOV

[MOPLAT-417]: https://artsyproduct.atlassian.net/browse/MOPLAT-417?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/screens/HomeTabs/Artists/Artists.tsx
```diff
@@ -23,6 +23,11 @@ export const Artists = () => {
 
   const items = zip(artists, counts)
 
+  navigation.navigate("ArtistTabs", {
```
patrinoua, line 26: let's remove this before merging

src/app/screens/HomeTabs/Artists/ArtistTabs/ArtistArtworks/ArtistArtworks.tsx
```diff
@@ -24,6 +24,11 @@ export const ArtistArtworks = ({ slug }: { slug: string }) => {
 
   const space = useSpace()
 
+  navigation.navigate("Artwork", {
```
patrinoua, line 27: let's remove this before merging

ios/Energy.xcodeproj/project.pbxproj
```diff
@@ -256,9 +286,7 @@
 						TestTargetID = 13B07F861A680F5B00A75B9A;
 					};
 					13B07F861A680F5B00A75B9A = {
-						DevelopmentTeam = 23KMWZ572J;
```
brainbicycle, line 259: can we make sure we untoggle automatic code signing in Xcode before merging 🙏

src/app/sharedUI/screens/Artwork/Artwork.tsx
```diff
@@ -81,3 +119,28 @@ export const Artwork = () => {
     </BottomSheetModalProvider>
   )
 }
+
+// eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
```
brainbicycle, line 123: **tangent, non-blocking:** I am seeing these pop up in my code as well, should we disable this rule? or is there a way to avoid it?

yarn.lock
```diff
@@ -4439,6 +5910,120 @@ expect@^28.1.0:
...
+
+expo-modules-core@0.9.2:
+  version "0.9.2"
+  resolved "https://registry.yarnpkg.com/expo-modules-core/-/expo-modules-core-0.9.2.tgz#657a3d804e73f3d41e6fa35d40a44aee5a4a287e"
+  integrity sha512-p/C0GJxFIIDGwmrWi70Q0ggfsgeUFS25ZkkBgoaHT7MVgiMjlKA/DCC3D6ZUkHl/JlzUm0aTftIGS8LWXsnZBw==
+  dependencies:
+    compare-versions "^3.4.0"
+    invariant "^2.2.4"
+
+expo@^45.0.0:
```
brainbicycle, line 5999: **thought, non-blocking**: it looks like using this expo dependency pulled in a ton of transitive dependencies and required a bunch of native code changes, I am wondering if there were any other non-expo email compose libraries that might be lighter? I don't know so much about expo so wondering if other have thoughts, maybe we will need it eventually anyway? @artsy/mobile-platform

src/app/sharedUI/screens/Artwork/Artwork.tsx
```diff
@@ -62,17 +88,29 @@ export const Artwork = () => {
...
-              navigateTo={() => {}}
+              onPress={() => {}}
             />
             <BottomSheetModalRow
               Icon={<EditIcon fill="onBackgroundHigh" />}
               label="Send by Email"
-              navigateTo={() => {}}
+              onPress={() => {
+                MailComposer.composeAsync({
+                  subject: `Information about "${title ? title : ""}" by ${
```
brainbicycle, line 98: **suggestion**: if there are no artistnames maybe we probably shouldn't show `by` either could add it to the ternary below

brainbicycle (changes requested):
some small change requests but looks good, larger question I have is about the expo dependency, it seems like a lot of requirements for an email compose library but maybe it is unavoidable, I will leave it to the rest of you who know more about expo. Nice work!

android/settings.gradle
```diff
@@ -1,3 +1,6 @@
 rootProject.name = 'Energy'
 apply from: file("../node_modules/@react-native-community/cli-platform-android/native_modules.gradle"); applyNativeModulesSettingsGradle(settings)
 include ':app'
+
+apply from: new File(["node", "--print", "require.resolve('expo/package.json')"].execute(null, rootDir).text.trim(), "../scripts/autolinking.gradle")
+useExpoModules()
```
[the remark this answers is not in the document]
pvinis: just for a new line at the end

ios/Energy.xcodeproj/project.pbxproj
pvinis, line 259: good point from @brainbicycle. we should edit this diff in this file.

changes that need to go are lines including:
- DevelopmentTeam
- ProvisioningStyle
- CODE_SIGN_STYLE
- PROVISIONING_PROFILE_SPECIFIER

everything else should stay. i think its basically these 4 lines. i can make gh suggestions if you like.

ios/Energy.xcodeproj/project.pbxproj
```diff
@@ -256,9 +286,7 @@
...
 						LastSwiftMigration = 1120;
```
pvinis, line 289:
```suggestion
						DevelopmentTeam = 23KMWZ572J;
						LastSwiftMigration = 1120;
						ProvisioningStyle = Manual;
```

src/app/sharedUI/screens/Artwork/Artwork.tsx
```diff
@@ -62,17 +88,29 @@ export const Artwork = () => {
...
+                    artistNames ? artistNames : ""
+                  }`,
+                  isHtml: true,
+                  body: bodyHTML,
```
pvinis, line 102: i know i shouldnt, but im impressed with how well this works :D

src/app/sharedUI/screens/Artwork/Artwork.tsx
pvinis: i guess this will come from the templates, right? so either we use the templates here, or we have a separate ticket and its ok to put whatever string here for now?

yarn.lock
admbtlr: Expo is a pretty huge dependency, and it can cause major headaches (although we've probably avoided the largest one, when they switched away from their unimodules system about 6 months ago). I would prefer not to use it unless there's no alternative.

brainbicycle: It is easy to revert now but if it is there we start to take it for granted and rely on it for other dependencies and then it gets harder 😬 if there is an easy other option I would prefer to use it personally

src/app/sharedUI/screens/Artwork/Artwork.tsx
pvinis: @patrinoua will make a separate PR for the templates, as part of the same ticket.

pvinis:
https://artsy.slack.com/archives/C02NAJ2CGLW/p1665073971939519 for the expo decision.

pvinis:
template to be worked on monday 👍

PR #264, opened 2022-10-12
fix: Bottom sheet margin fix

rajsam003:
|Before|After|
|---|---|
|image|image|

### PR Checklist

### To the reviewers 👀

[docs]: /docs/README.md


pvinis (approved):
this feels like a weird way to solve this but lets merge.
i would expect that there would be a safearea thing for the view in the modal.

PR #271, opened 2022-10-18
fix: xcode build

pvinis:
### Description

This PR resolves []

This PR is fixing the xcode build. I was having issues building the app, this fixed it.

### PR Checklist

### To the reviewers 👀

[docs]: /docs/README.md


package.json
```diff
@@ -46,8 +46,8 @@
     "easy-peasy": "^5.0.4",
     "events": "^3.3.0",
     "expect": "^28.1.0",
-    "expo": "^45.0.0",
-    "expo-mail-composer": "~11.2.0",
+    "expo": "^46.0.16",
```
pvinis, line 49: comes from https://github.com/expo/expo/issues/18490#issuecomment-1223394351

patches/react-native+0.66.2.patch
```diff
@@ -0,0 +1,17 @@
+diff --git a/node_modules/react-native/scripts/find-node.sh b/node_modules/react-native/scripts/find-node.sh
+index c498ba6..85f4d6b 100755
+--- a/node_modules/react-native/scripts/find-node.sh
++++ b/node_modules/react-native/scripts/find-node.sh
+@@ -37,3 +37,12 @@ if [[ ! -x node && -d ${HOME}/.anyenv/bin ]]; then
+     eval "$(anyenv init -)"
+   fi
+ fi
++
++# Set up asdf-vm if present
```
pvinis, line 10: this is needed since we (i) have asdf, and RN 66 doesn't support it. we should get to later versions, and we can remove that right after 👍

rajsam003: Do we have any other alternatives?

PR #312, opened 2022-11-09
fix : Artwork border box and About the artwork title should be hidden if artwork detail is empty

rajsam003:
This PR resolves [MOPLAT-495]

### PR Checklist

### To the reviewers 👀

[docs]: /docs/README.md

[MOPLAT-495]: https://artsyproduct.atlassian.net/browse/MOPLAT-495?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/sharedUI/screens/Artwork/ArtworkContent/ArtworkContent.tsx
```diff
@@ -130,7 +130,6 @@ export const ArtworkContent = ({ slug }: { slug: string }) => {
     imageRights ||
     inventoryId ||
     confidentialNotes ||
-    internalDisplayPrice ||
```
pvinis, line 133: this feels.. weird? why did it now show with that line? was it `""`?

also, would it show the box when there is only an internal display price now, or will we figure out the box is missing in that specific case? 🤔

pvinis (changes requested):
not sure about this change.

src/app/sharedUI/screens/Artwork/ArtworkContent/ArtworkContent.tsx
rajsam003: I basically just removed this from `shouldDisplayTheDetailBox` as we are not displaying the price inside that box. We are displaying `internalDisplayPrice ` and `price` above the box

pvinis (approved):

PR #358, opened 2022-12-12
chore(relay): Upgrade to Relay 14

damassi:
### Description

This upgrades Energy to Relay 14 and also simplifies some naming in a common function to remove unnecessary words. 

cc @artsy/mobile-platform


src/app/screens/HomeTabs/Artists/ArtistTabs/ArtistArtworks/ArtistArtworks.tests.tsx
```diff
@@ -1,6 +1,6 @@
 import { range } from "lodash"
 import { __globalStoreTestUtils__ } from "app/store/GlobalStore"
-import { mockEnvironmentPayloadMaybe } from "shared/tests/mockEnvironmentPayload"
+import { mockEnvironmentPayload } from "shared/tests/mockEnvironmentPayload"
```
damassi, line 3: Since this function is used all over there's no reason to make the function name longer than necessary (as well as diverge from the file name itself)

scripts/setup-env-for-artsy
```diff
@@ -1,6 +1,8 @@
 #!/usr/bin/env bash
 set -euxo pipefail
 
+# Create Relay generated folder
+mkdir src/__generated__
```
damassi, line 5: Someone needs to PR folder creation to the `relay-compiler`! Once that's in we can remove these create steps.

.circleci/config.yml
```diff
@@ -66,6 +66,10 @@ commands:
             - node_modules
   run-relay-compiler:
     steps:
+      - run:
+          name: Create relay `__generated__` folder
+          command: |
+            mkdir src/__generated__
```
pvinis, line 72: lets remove this, and add a `__generated__/.gitkeep` file. makes things simpler

src/shared/tests/mockEnvironmentPayload.ts
```diff
@@ -29,16 +29,9 @@ const generateID = (pathComponents: readonly string[] | undefined) => {
...
-interface MockResolverContext {
-  parentType?: string | undefined
-  name?: string | undefined
-  alias?: string | undefined
-  path?: ReadonlyArray<string> | undefined
-  args?: { [key: string]: any } | undefined
-}
-
 let paths: { [name: string]: string } = {}
-const goodMockResolver = (ctx: MockResolverContext) => {
```
pvinis, line 41: ah, its not good anymore :p

src/app/screens/HomeTabs/Artists/ArtistTabs/ArtistArtworks/ArtistArtworks.tests.tsx
pvinis: i really dont remember what the `maybe` was for.. but, makes sense, lets keep it simple.

scripts/setup-env-for-artsy
pvinis: same suggestion as below, then we dont need this special treatment of this folder.

damassi: ah nice, i didn't realize that a .gitkeep would still override the gitignore 👍

src/shared/tests/mockEnvironmentPayload.ts
damassi: its still good 😄

pvinis (approved):

PR #374, opened 2022-12-16
feat: offline images

pvinis:
This PR resolves [MOPLAT-]

### Description
- the images use react-native-fs to get cached.
- we have `CachedImage` that will take a url (usually from gemini) and if its cached it will display the cached one, or if its not it will display the remote one. *it will not do any caching if it displays the cached one, that's for a later PR.*
- for images, we use a `urlMap`, from an image url (like `https://weriouwoeiur.artsy.net/whatever/image.jpg&size=1000`) to a uuid (like `2342-abc2-oiu-oiu`).

### Next work
- make sure documents also cache
- cache things if we are online, to keep the cache fresh

### PR Checklist

### To the reviewers 👀

[docs]: /docs/README.md


src/app/sharedUI/items/AlbumListImage.tsx
```diff
@@ -15,15 +16,14 @@ export const AlbumListImage = ({ slug, style }: AlbumListImageProps) => {
   const placeHolderHeight = useScreenDimensions().height / 5
   const albumListImage = albumImages.artwork?.image
 
+  const uri = useCachedOrFetchUrl(albumListImage?.resized?.url ?? "")
```
damassi, line 19: Will we need to wrap all images like this going forward?

src/app/sharedUI/items/AlbumListImage.tsx
```diff
@@ -15,15 +16,14 @@ export const AlbumListImage = ({ slug, style }: AlbumListImageProps) => {
...
   if (!albumListImage?.resized?.url) {
     return <ImagePlaceholder height={placeHolderHeight} />
   }
 
   return (
-    <Image
-      source={{ uri: albumListImage.resized.url }}
-      style={[style, { aspectRatio: albumListImage?.aspectRatio ?? 1 }]}
-    />
+    <Image source={{ uri }} style={[style, { aspectRatio: albumListImage?.aspectRatio ?? 1 }]} />
```
damassi, line 26: Rather than wrapping the `uri` ref, can we create some kind of `<SystemImage />` component that handles this internally? (Trying to think of a better name 🤔)

src/app/sharedUI/items/ArtworkGridItem.tsx
```diff
@@ -30,11 +31,13 @@ export const ArtworkGridItem = ({
     (state) => state.presentationMode.hiddenItems.worksAvailability
   )
 
+  const uri = useCachedOrFetchUrl(artwork.image?.resized?.url ?? "")
```
damassi, line 34: ditto (and anywhere else)

src/app/sharedUI/items/DocumentGridItem.tsx
```diff
@@ -25,70 +25,28 @@ interface DocumentGridItemProps {
...
   const fileExtension = last(document.url.split("."))
-  const userAccessToken = GlobalStore.useAppState((state) => state.auth.userAccessToken)!
   const isSelectModeActive = GlobalStore.useAppState((state) => state.selectMode.isActive)
 
-  const downloadFile = async (fileUrl: string, dest: string) => {
-    try {
-      setIsDownloading(true)
-      const fetchInstance = RNFetchBlob.config({ path: dest })
+  const url = useCachedOrFetchUrl(document.url)
+  let realUrl = url
```
damassi, line 32: Is this needed?

src/app/system/sync/fileCache.ts
```diff
@@ -0,0 +1,201 @@
...
+const PATH_CACHE_DOCUMENTS = PATH_CACHE + "/documents"
+const PATH_CACHE_RELAY_DATA = PATH_CACHE + "/relayData"
+
+/**
+ * This is just for debugging!
+ * The app doesnt need an extension here.
+ * But if we want to debug and see the images in the simulator's file system in Finder,
+ * then adding an extension will make quicklook show them as images.
+ * eg. If you go to `~/Library/Developer/CoreSimulator/Devices/uuid-of-simulator/data/Containers/Data/Application/uuid-of-application/Documents/cache/images`
+ * you can see the cached images in there.
```
damassi, line 24: 👍

src/app/system/sync/fileCache.ts
```diff
@@ -0,0 +1,201 @@
...
+        return
+      }
+    }
+    doIt()
+  }, [])
+
+  return pathOrUrl
+}
+
+let urlMap: Record<string, string> = {}
```
damassi, line 184: (nonblocking): Can we move this closer to the top of the file with the rest of the file-local variables?

src/app/system/sync/utils/mapAsync.ts
```diff
@@ -0,0 +1,24 @@
...
+}
+
+export const forEachAsyncLinear = async <T>(
+  items: T[],
+  callback: (item: T) => Promise<unknown>
+) => {
+  for (const item of items) {
+    await callback(item)
+  }
+}
```
damassi, line 24: If any of these functions aren't being used yet can we delete? Or if we're certain we're going to use them soon can we rename this file something like `iterators.ts`?

damassi (approved):
Only a couple minor comments from me! Nice and simple implementation with the `fileMap` and the single function to retrieve things 👍 

There are a handful of quick design suggestions below that I think could simplify the implementation but we can also cleanup / refactor later if necessary. Nothing blocking! 

Only nice-to-have in here request would be to create a new `<SystemImage />` component wrapper that imports the file uri retriever hook and then update the `Dangerfile.ts` to guard against future `<Image />` imports. Then devs can just use that and never have to worry about additional implementation details, and we never have to worry about misses.

src/app/system/sync/fileCache.ts
```diff
@@ -0,0 +1,201 @@
...
+  })()
+
+  return path + `/${filename}`
+}
+
+export const useCachedOrFetchUrl = (url: string): string => {
+  const [pathOrUrl, setPathOrUrl] = useState(url)
+
+  useEffect(() => {
+    const doIt = async () => {
```
damassi, line 157: #superminor - I'm no fun when it comes to naming stuff even if its harmless 😄 Just because code always changes and grows in time and something thats a simple executor right now could later be more involved (or cut and pasted elsewhere, etc). So yah, more descriptive is always better IMO. 

Maybe something like
```tsx
const retrievePath = async () => { ... }
```

src/app/navigation/Main.tsx
```diff
@@ -94,7 +95,14 @@ export const Main = () => {
   const isDarkMode = GlobalStore.useAppState((s) => s.devicePrefs.colorScheme === "dark")
 
   useEffect(() => {
-    if (isRehydrated) SplashScreen.hide()
+    const doIt = async () => {
```
damassi, line 98: See https://github.com/artsy/energy/pull/374/files#r1051294570

src/app/system/sync/fileCache.ts
```diff
@@ -0,0 +1,201 @@
...
+  } catch (error) {
+    console.log("[fileCache] Error saving file:", error)
+  }
+}
+
+export const downloadImageToCache = async (url: string) => {
+  const id = uuidv4()
+  const filePath = getFilePath({ type: "image", filename: id + IMAGE_EXTENSION })
+
+  await makeSureCacheIsReady()
```
damassi, line 60: We'll wanna wrap all of this in a try catch or otherwise guard

src/app/system/sync/fileCache.ts
```diff
@@ -0,0 +1,201 @@
...
+}
+
+export const downloadDocumentToCache = async (url: string): Promise<string> => {
+  const id = uuidv4()
+  const fileExtension = url.split(".").pop()
+  const filePath = getFilePath({ type: "document", filename: id + "." + fileExtension })
+
+  console.log("downloading file", { url, filePath })
+
+  await makeSureCacheIsReady()
```
damassi, line 82: try/catch

src/app/sharedUI/items/AlbumListImage.tsx
damassi: We can also guard against the undefined case inside of the function rather than outside of it

src/app/sharedUI/items/DocumentGridItem.tsx
```diff
@@ -25,70 +25,28 @@ interface DocumentGridItemProps {
...
-        return
-      }
+    const isFileCached = url.startsWith("file://")
 
-      console.error(error)
+    if (!isFileCached) {
+      setIsDownloading(true)
+      realUrl = await downloadDocumentToCache(document.url)
+      console.log({ url2: realUrl })
+      await saveUrlMap()
```
damassi, line 43: Can this be called from inside `downloadDocumentToCache`?

src/app/system/sync/fileCache.ts
```diff
@@ -0,0 +1,201 @@
...
+  path?: string
+}
+
+interface SaveFileProps extends FileProps {
+  data: string
+}
+
+export const saveFileToCache = async ({ data, filename, type, path }: SaveFileProps) => {
+  if (type === undefined && path === undefined) {
+    console.log("[fileCache] Error saving file: `type` or `path` must be defined.")
```
damassi, line 42:
```suggestion
    console.log("[fileCache] Error saving file: `type` and `path` must be defined.")
```

src/app/system/sync/fileCache.ts
```diff
@@ -0,0 +1,201 @@
...
+  type?: "image" | "document" | "relayData"
+  path?: string
+}
+
+interface SaveFileProps extends FileProps {
+  data: string
+}
+
+export const saveFileToCache = async ({ data, filename, type, path }: SaveFileProps) => {
+  if (type === undefined && path === undefined) {
```
damassi, line 41:
```suggestion
  if (!type && !path) {
```

src/app/system/sync/fileCache.ts
```diff
@@ -0,0 +1,201 @@
...
+    console.log("[fileCache] Error saving file: `type` or `path` must be defined.")
+    return
+  }
+
+  const actualPath = type !== undefined ? getFilePath({ type, filename }) : path!
```
damassi, line 46:
```suggestion
  const actualPath = type ? getFilePath({ type, filename }) : path!
```

src/app/system/sync/fileCache.ts
```diff
@@ -0,0 +1,201 @@
...
+  const id = uuidv4()
+  const filePath = getFilePath({ type: "image", filename: id + IMAGE_EXTENSION })
+
+  await makeSureCacheIsReady()
+  await downloadFile({
+    fromUrl: url,
+    toFile: filePath,
+  })
+
+  if (urlMap[url] !== undefined) {
```
damassi, line 66:
```suggestion
  const urlExists = !!urlMap[url]
  if (urlExists) {
```

src/app/system/sync/fileCache.ts
damassi: We'll also want to move this to the top of each function where its used so that we can warm stuff up or bail/notify early before executing any further operations

src/app/system/sync/fileCache.ts
```diff
@@ -0,0 +1,201 @@
...
+    try {
+      await unlink(filePath)
+    } catch (e) {}
+  }
+
+  urlMap[url] = id + "." + fileExtension
+  return filePath
+}
```
damassi, line 96: The `downloadImageToCache` and `downloadDocumentToCache` are pretty much identical; can we combine the two and use an object argument for the inputs?

src/app/system/sync/fileCache.ts
```diff
@@ -0,0 +1,201 @@
...
+        return PATH_CACHE_IMAGES
+      case "relayData":
+        return PATH_CACHE_RELAY_DATA
+    }
+  })()
+
+  return path + `/${filename}`
+}
+
+export const useCachedOrFetchUrl = (url: string): string => {
```
damassi, line 153: Wondering.. Are we treating this as a generalized file cache to make things snappier on future boots, or only thinking about it in terms of offline? 

It kinda makes sense at this point to just store all images inside of a cache no matter what, so that when a user browses around it builds things up over time... So something like

```tsx
export const useUrlCache = ({ type, url }): string => {
    const path = type === "image" ? 
      await downloadImageToCache(url) : 
      await downloadDocumentToCache(url) 
    
    return path 
}
```
And then internal to `downloadImageToCache` and `downloadDocumentToCache` we can perform all of the operations below. 

Alternatively, if we're _only_ treating this as a cache for those who have sync'd offline, we should probably bail early and just return `url` by checking against 
```tsx
const isOnline = GlobalStore.useState(state => state.network.isOnline)
```

But yeah, the "always on" file cache seems easy at this point.

src/app/system/sync/fileCache.ts
damassi: If this more verbose `undefined` form is deliberate, no prob keeping here and below! (just calling out that its not necessary since we're not mixing `null` and `undefined` anywhere; `!truthy` is sufficient)

src/app/store/Models/NetworkStatusModel.ts
```diff
@@ -0,0 +1,38 @@
...
+  fetchPolicy: "store-only",
+
+  toggleNetworkConnection: action((state, isOnline) => {
+    state.isOnline = isOnline
+  }),
+
+  determineFetchPolicy: action((state) => {
+    if (state.isOnline) {
+      console.log("ONLINE")
+      state.fetchPolicy = "network-only"
```
damassi, line 28: Don't need to update this here, but thinking the default should be `store-or-network` if online, that way pages user has visited will be cached on subsequent visits (which would probably happen a lot at a fair)

src/app/sharedUI/items/AlbumListImage.tsx
pvinis: do you think that is better? i was 50/50 on a hook with regular Image, or no hook and custom Image.

src/app/sharedUI/items/DocumentGridItem.tsx
pvinis: this line is me trying to debug stuff.

but the line above, is a reason why wrapping Image might be a bad idea. We don't use this URL here for an Image, we use it to give it to the FileViewer below. If we wrap Image, that means that we also need to wrap FileViewer and anything else, which feels like too much. Getting a url from a hook feels more general.

what do you think?

src/app/store/Models/NetworkStatusModel.ts
pvinis: this is all your code :D im rebasing

src/app/system/sync/fileCache.ts
pvinis: sufficient, yes. readable, questionable 😅.

src/app/system/sync/fileCache.ts
pvinis: `or` is right. if we have type, we can guess the path. if we have path, we have path. we don't need both, hence the if to check if both are undef.

src/app/system/sync/fileCache.ts
pvinis: will do

src/app/system/sync/fileCache.ts
pvinis: will do

src/app/system/sync/utils/mapAsync.ts
pvinis: already used. renamed.

src/app/wrappers/ErrorBoundary.tsx
```diff
@@ -19,11 +21,16 @@ export class ErrorBoundary extends Component<ErrorBoundaryProps, ErrorBoundarySt
...
 
   render() {
-    const { children, withoutBackButton } = this.props
+    const { children, withoutBackButton, Fallback } = this.props
     const { error } = this.state
 
     if (error) {
       this.props.catch?.(error)
+
+      if (Fallback !== undefined) {
```
pvinis, line 30: adding a little extra fallback here, to reuse this error boundary comp

src/app/sharedUI/items/DocumentGridItem.tsx
pvinis: yep. it was, ill change it back. something was messing with it and I tried to move it outside, but it didn't change anything, so it wasn't that. i just pushed while I was at this debugging state hehe

src/app/sharedUI/items/AlbumListImage.tsx
pvinis: i changed the hook to take undef too, but to throw if it does. then I added an errorboundary around that component. feels better?

src/app/App.tsx
```diff
@@ -7,10 +7,8 @@ setupFlipper()
 
 LogBox.ignoreLogs(["lineHeight", "borderRadius", "fontSize", "onBackground", "borderWidth"])
 
-export const App = () => {
-  return (
```
pvinis, line 11: keeping this that way will make it less likely people will try to add hooks here. we should avoid adding hooks here.

src/app/sharedUI/items/DocumentGridItem.tsx
pvinis: ok so this is one of the things to-be-done.

last I tried, on a document, I would tap and it would download the file, I could see it in the filesystem, but for some reason it would not open the local URL. FileViewer would still get the https URL. i was messing around trying to fix this, but I'm sure its not needed to have this extra realurl or whatever. see if you get the same by just removing the realurl and using just url.

src/app/sharedUI/items/DocumentGridItem.tsx
pvinis: yea, this should go back inside the downloadfiletocache

src/app/system/sync/fileCache.ts
```diff
@@ -100,12 +134,62 @@ const getFilePath = ({ filename, type }: FileProps) => {
...
     }
   })()
 
   return path + `/${filename}`
 }
 
-const urlMap: Record<string, string> = {} // export as json
+export const useCachedOrFetchUrl = (url: string | undefined): string => {
+  if (url === undefined) {
+    throw new Error("url is undefined")
```
pvinis, line 147: thats the new little thing. see how you feel about this, and wrapping stuff with errorboundaries.

src/app/system/sync/syncManager.ts
```diff
@@ -261,6 +262,42 @@ export function initSyncManager({
     updateStatus("Complete. `partnerShowTabsQuery`", syncResults.partnerShowTabsQuery)
   }
 
+  const syncImages = async () => {
+    updateStatus("Syncing images...")
+
+    const urls = parsers.getImageUrls()
+
+    await forEachAsyncLinear(urls, async (url) => {
```
pvinis, line 270: these should be changed to `forEachAsync`, and remove the linear bit. i used it here to debug.

src/app/system/sync/syncManager.ts
```diff
@@ -261,6 +262,42 @@ export function initSyncManager({
...
+      await downloadFileToCache({ url, type: "image" })
+    })
+  }
+
+  const syncInstallShots = async () => {
+    updateStatus("Syncing install shots...")
+
+    const urls = parsers.getInstallShotUrls()
+
+    await forEachAsyncLinear(urls, async (url) => {
```
pvinis, line 280: same

src/app/system/sync/fileCache.ts
```diff
@@ -69,11 +102,12 @@ export const getFileFromCache = async ({ filename, type }: FileProps) => {
 
 const makeSureCacheIsReady = async () => {
   try {
+    // console.log("cache location", PATH_CACHE)
```
pvinis, line 105: ⚠️ uncomment this if you wanna get to the filesystem of the simulator. just get the path, and run `open <path>` and you're there.

src/app/system/sync/fileCache.ts
damassi: 😄 no prob 👍

src/app/App.tsx
damassi: Oh I see -- Ha, i think this is the first legit argument for an implicit return i've heard. Makes sense 👍

src/app/sharedUI/items/AlbumListItem.tsx
```diff
@@ -19,14 +23,16 @@ export const AlbumListItem = ({ album }: { album: Album }) => {
...
         {first3Artworks.reverse().map((artworkId) => (
-          <AlbumListImage
-            slug={artworkId}
-            key={artworkId}
-            style={{
-              flex: 1,
-              marginRight: -overlapSize,
-            }}
-          />
+          <ErrorBoundary Fallback={<ImagePlaceholder height={placeHolderHeight} />}>
```
damassi, line 26: I like! This is still an implementation detail that future devs shouldn't need to remember tho; will move it into the image component and provide a prop

src/app/system/sync/fileCache.ts
damassi: I like this a lot! Better to have nice fallbacks and a papertrail then just weird empty spots

src/app/system/sync/syncManager.ts
```diff
@@ -330,6 +367,31 @@ const parsers = {
...
     return showSlugs
   },
+
+  getImageUrls: (): string[] => {
+    return (
+      syncResults.artworkContentQuery?.map(
+        (artworkContent) => artworkContent.artwork?.image?.resized?.url ?? ""
+      ) ?? []
+    ).filter((url) => url !== "")
+    // TODO: Are these all the images?
```
damassi, line 377: I believe these are all the images! I was digging through and debugging at some point and everything seemed to map up 1:1 from the artist list

src/app/system/sync/syncManager.ts
```diff
@@ -346,14 +408,14 @@ const persistDataToOfflineCache = async (relayEnvironment: RelayModernEnvironmen
   await saveFileToCache({
     data: JSON.stringify(relayData),
```
damassi, line 409: Woke up this morning with the thought that i'm stringifying this unnecessarily -- we can just save it as straight JSON and skip the parsing back and forth. (Thinking here was that this was the web, where one can't pass anything but stringified json from server to client on the rehydration step)

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true">Fails</th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  Please use `CachedImage` instead of `Image`. This enables offline image viewing support. See:
> `src/app/wrappers/CachedImage.tsx`
  </td>
    </tr>
  </tbody>
</table>




<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 983c4e3bc393088f0ca16af0aa3e57fcd93d32c9
</p>

src/app/system/sync/syncManager.ts
```diff
@@ -265,35 +266,51 @@ export function initSyncManager({
...
+   */
+
   const syncImages = async () => {
     updateStatus("Syncing images...")
 
     const urls = parsers.getImageUrls()
 
-    await forEachAsyncLinear(urls, async (url) => {
-      await downloadFileToCache({ url, type: "image" })
+    const imageUrls = await mapAsync(urls, async (url) => {
```
damassi, line 278: me just debugging stuff

src/app/system/sync/syncManager.ts
```diff
@@ -265,35 +266,51 @@ export function initSyncManager({
...
+    console.log(imageUrls)
   }
 
   const syncInstallShots = async () => {
     updateStatus("Syncing install shots...")
 
     const urls = parsers.getInstallShotUrls()
 
-    await forEachAsyncLinear(urls, async (url) => {
+    const localURLS = await mapAsync(urls, async (url) => {
```
damassi, line 292: same here (and so on)

src/app/system/sync/utils/asyncIterators.ts
```diff
@@ -1,24 +1,34 @@
...
-export const forEachAsync = async <T>(items: T[], callback: (item: T) => Promise<any>) => {
+export const forEachAsync = async <T>(
+  items: T[],
+  callback: (item: T, index: number) => Promise<any>
+) => {
   await Promise.all(
-    items.map(async (item) => {
-      await callback(item)
+    items.map((item, index) => {
+      callback(item, index)
```
damassi, line 18: @pvinis - i removed the async/await from here because this will force things into sync mode, vs iterating through and just returning promises to be executed in the Promise.all internal iterator

damassi: I could be missing something silly tho... is my thinking right here?

src/app/screens/HomeTabs/Artists/ArtistTabs/ArtistShows/ArtistShows.tsx
```diff
@@ -62,9 +62,14 @@ export const artistShowsQuery = graphql`
...
           node {
+            ...ShowListItem_show @arguments(imageSize: $imageSize)
             internalID
             slug
-            ...ShowListItem_show @arguments(imageSize: $imageSize)
+            coverImage {
+              resized(width: $imageSize, version: "normalized") {
+                url
+              }
+            }
```
damassi, line 72: This is kinda unfortunate but not sure how to dig in here from within the sync manager. The issue is that you can't do `@relay(mask: false)` on fragments that have arguments 😢

src/app/screens/HomeTabs/Artists/ArtistTabs/ArtistShows/ArtistShows.tsx
```diff
@@ -57,7 +57,7 @@ export const ArtistShows = ({ slug }: { slug: string }) => {
 }
 
 export const artistShowsQuery = graphql`
-  query ArtistShowsQuery($partnerID: String!, $slug: String!, $imageSize: Int!) {
+  query ArtistShowsQuery($partnerID: String!, $slug: String!, $imageSize: Int!) @raw_response_type {
```
damassi, line 60: This `@raw_response_type` will generate the correct type tho, which is nice. But doesn't help us actually access raw data

src/app/system/sync/utils/asyncIterators.ts
pvinis: what i was trying to avoid/make clear is the fact that this is an async bit, and the callback is an async thing too, so that no one goes and adds something here and mess up the async-icity of the whole helper. but I get your point, so lets keep it this way 👍

src/app/system/sync/fileCache.ts
```diff
@@ -8,51 +8,101 @@ import {
...
+      console.log("[fileCache] Error saving file: `type` or `path` must be defined.")
+      return
+    }
 
-  const path = getFilePath({ type, filename })
+    await makeSureCacheIsReady()
 
-  try {
-    await writeFile(path, data, "utf8")
+    const actualPath = type !== undefined ? getFilePath({ type, filename }) : path!
```
damassi, line 54: `const filePath =`

src/app/system/sync/fileCache.ts
```diff
@@ -8,51 +8,101 @@ import {
...
+    await makeSureCacheIsReady()
+
+    const id = uuidv4()
 
-  const id = nanoid()
+    const extension = (() => {
+      switch (type) {
+        case "image":
+          return IMAGE_EXTENSION
+        case "document":
```
damassi, line 78: Not sure if folio might have other types of documents in the future so turned these listy things into a switch, I might be overdoing it a bit for now but not so terrible

src/app/system/sync/fileCache.ts
damassi: Sounds good 👍

src/app/system/sync/syncManager.ts
```diff
@@ -260,6 +264,49 @@ export function initSyncManager({
...
+      })
+    )
+  }
+
+  const syncDocuments = async () => {
+    updateStatus("Syncing documents...")
+
+    const urls = parsers.getDocumentsUrls()
+
+    await mapAsync(urls, (url) =>
```
damassi, line 302: `forEachAsync`

src/app/system/sync/syncManager.ts
```diff
@@ -345,14 +419,14 @@ const persistDataToOfflineCache = async (relayEnvironment: RelayModernEnvironmen
   await saveFileToCache({
     data: JSON.stringify(relayData),
     filename: "relayData.json",
-    type: "json",
+    type: "relayData",
```
damassi, line 422: 👍

src/app/system/sync/fileCache.ts
pvinis: i felt it was ok with just the one "special" case of docs, but I guess lets start this way and we see

src/app/system/sync/syncManager.ts
pvinis: 👍

damassi (approved):
Looks great 👍

PR #377, opened 2022-12-16
refactor(relay-network): Make network initialization deterministic

damassi:
### Description

When refactoring the relay network to support offline caching noticed that there were lots of errors related to the global store not being initialized, due to the network layer being created statically at runtime. This PR 
- Removes the use of a static, globally available `defaultNetwork` in favor of accessing the relay network directly from the context
- Removes test-level mocking of non-test network imports, and explicitly sets up mock test environments for our tests

cc @artsy/mobile-platform 

### PR Checklist

### To the reviewers 👀

[docs]: /docs/README.md


src/app/relay/environment/createEnvironment.ts
```diff
@@ -1,42 +1,55 @@
...
-    // See: https://github.com/relay-tools/react-relay-network-modern#advanced-options-2nd-argument-after-middlewares
-    noThrow: true,
-  }
-) // as second arg you may pass advanced options for RRNL
+const createNetworkLayer = () => {
+  /**
+   * FIXME: As the following rely on the global store, we can't be sure that
+   * it's available at import time and so the below have been moved to lazy
+   * imports.
+   */
```
MounirDhahri, line 16: **thought:** have you tried initiating the relay environment only after the store is done rehydrating within `RelayProvider.tsx`

MounirDhahri (commented):
nice changes

src/app/relay/environment/createEnvironment.ts
damassi: I did try this but recall there still being some subtle race conditions. I want to return to this after this sprint!

PR #382, opened 2022-12-19
chore(eslint): Add rules-of-hooks linting

damassi:
### Description

After convo with @pvinis realized we didn't have the rules-of-hooks eslint plugin configured. So this fixes that. 

cc @artsy/mobile-platform 

### PR Checklist

### To the reviewers 👀

[docs]: /docs/README.md


src/app/screens/HomeTabs/Albums/CreateOrEditAlbum/useArtworksByMode.ts
```diff
@@ -2,7 +2,7 @@ import { GlobalStore } from "app/store/GlobalStore"
 import { AlbumsModel } from "app/store/Models/AlbumsModel"
 
 const getSelectedArtworksFromAlbum = (type: keyof AlbumsModel["sessionState"], slug?: string) => {
-  const artworks = GlobalStore.useAppState((state) => state.albums.sessionState[type])
```
pvinis, line 5: hmmm is this change because of the name of the function?

should we make this a hook officially and keep the hook here?

pvinis (approved):
one thing, otherwise all good. thank you!

src/app/screens/HomeTabs/Albums/CreateOrEditAlbum/useArtworksByMode.ts
damassi: Yep because the name of the function 👍 I was thinking about changing it to a hook but below, where its used there's conditional logic, so it would break.

PR #391, opened 2022-12-21
feat: add documents in the offline cache

pvinis:
This PR resolves [MOPLAT-]

### Description

docs now work, with and without cache. they also use the right name, so we can share/email correctly.

i also added a little debug view for us to see whats up with the cache, as well as two small UI bugs.

Screenshot 2022-12-22 at 23 55 36

### PR Checklist

### To the reviewers 👀

[docs]: /docs/README.md


src/palette/organisms/Screen/exposed/Body.tsx
```diff
@@ -56,7 +57,7 @@ export const Body = ({
           <ArtsyKeyboardAvoidingView>
             <ScrollView
               keyboardShouldPersistTaps="handled"
-              contentInset={{ bottom: bottomViewHeight - insets.bottom }}
```
pvinis, line 59: this was wrong. we just need a bit of space, and that's it. nothing to do with safe area

src/global.d.ts
```diff
@@ -3,7 +3,3 @@ import "jest-extended"
 declare global {
   const __TEST__: boolean
 }
-
-type DeepPartial<T extends object> = Partial<{
```
pvinis, line 7: moved so we dont have to import

src/app/utils/assertNever.ts
```diff
@@ -1,3 +0,0 @@
-export function assertNever(x: never): never {
```
pvinis, line 1: also moved so we dont have to import

src/app/system/sync/fileCache.ts
```diff
@@ -62,29 +68,40 @@ export const saveFileToCache = async ({ data, filename, type, path }: SaveFilePr
...
 
-    const extension = (() => {
+    const filename = (() => {
       switch (type) {
         case "image":
-          return IMAGE_EXTENSION
+          return id + IMAGE_EXTENSION
         case "document":
-          return "." + url.split(".").pop()
+          return url.split("/").pop()!
```
pvinis, line 87: documents use the actual filename from the url

src/app/screens/HomeTabs/Settings/Settings.tsx
```diff
@@ -181,6 +181,21 @@ export const Settings = () => {
...
+
+        <Button
+          block
+          onPress={() => {
+            GlobalStore.actions.networkStatus.toggleConnected(!isOnline)
+          }}
+        >
+          Is Online: {isOnline ? "true" : "false"}
+        </Button>
+
```
pvinis, line 198:
```suggestion
```
i guess that was removed

src/app/screens/Dev/BrowseOfflineCache.tsx
```diff
@@ -0,0 +1,44 @@
+import { Text } from "@artsy/palette-mobile"
+import { useEffect, useState } from "react"
+import { isTreeDirectory, tree, TreeNode } from "app/system/sync/fileCache"
+import { Screen } from "palette"
+
+export const BrowseOfflineCache = () => {
```
pvinis, line 6: magical

src/app/components/Items/DocumentGridItem.tsx
```diff
@@ -22,29 +22,37 @@ interface DocumentGridItemProps {
 }
 
 export const DocumentGridItem = ({ document, selectedToAdd, onPress }: DocumentGridItemProps) => {
+  const [isOpening, setIsOpening] = useState(false)
```
pvinis, line 25: this logic is a bit funky, but actually very simple, and the best I could think of.

init vals:
- opening: false
- downloading: false
We tap a document. opening goes true, because we are trying to open the doc. it looks for a local uri. if we have one, then it just opens it, and sets the opening to false, because we have opened it. if we don't have a local uri, then it will download it first, and then continue the same for opening. so in the useEffect, if we are not downloading (so we have the local uri) and we are opening, then we open the file with the fileviewer.

src/app/components/AddItemsToAlbum.tsx
```diff
@@ -114,10 +116,7 @@ export const AddItemsToAlbum = () => {
               </Flex>
             )
           }}
-          style={{
-            top: space(2),
-            marginBottom: space(12),
```
pvinis, line 119: this has to do with the bottom view, so we use that instead of a random number.

src/app/Providers.tsx
```diff
@@ -19,6 +20,7 @@ export const AppProviders = ({ children }: { children: React.ReactNode }) => {
                 <ProvideScreenDimensions>
                   {/*  */}
                   {children}
+                  <JotaiNexus />
```
pvinis, line 23: thats a tiny little lib that allows us to read/write the jotai state. used for the urlMap for cache.

src/app/components/AddItemsToAlbum.tsx
```diff
@@ -7,11 +7,13 @@ import { AlbumListItem } from "app/components/Items/AlbumListItem"
 import { useNavigationSavedForKey } from "app/system/hooks/useNavigationSave"
 import { GlobalStore } from "app/system/store/GlobalStore"
 import { Screen } from "palette"
+import { useScreenBottomViewHeight } from "palette/organisms/Screen/atoms"
```
damassi, line 10: This is more of a meta comment that we should follow up with; The use of atoms is cool as functionality, but would rather not encode more concepts into the file or folder tree and instead utilize these things transparently. If we're reading the `ScreenBottomViewHeight`, a developer doesn't need to know anything about an "atom" or what it might mean; they could simply 
```tsx
import { useScreenBottomViewHeight } from "palette/organisms/Screen"
``` 
or even better
```
import { useScreenBottomViewHeight } from "palette/hooks/useScreenBottomViewHeight"
```

They want the screen height, they use the hook. Less information in the system, less concepts to deal with, same functionality. 

Also, `atom` is a term that's being overloaded in two separate dimensions here -- it's an 'atom' (state concept), however it's nested in a folder called `organisms`. It's _also_ a value, which is an atom in terms of atomic design. This is really confusing and can be avoided completely by just leveraging every-day literate code styles.

src/app/screens/Dev/BrowseOfflineCache.tsx
```diff
@@ -0,0 +1,44 @@
...
+          <Text key={path} numberOfLines={1} ellipsizeMode="middle">
+            {"__".repeat(depth)}
+            {path.split("/").pop()!} <Text color="brand">{files[path].prettySize}</Text>
+          </Text>
+          {subtree}
+        </>
+      )
+    })}
+  </>
+)
```
damassi, line 44: Amazing 👌

src/app/system/sync/fileCache.ts
```diff
@@ -212,11 +231,48 @@ export const loadUrlMap = async () => {
...
+      data: JSON.stringify(readAtom(urlMapAtom)),
       filename: "urlMap",
       path: PATH_CACHE + "/urlMap.json",
     })
   } catch (error) {
     console.log("[fileCache] Error saving urlMap:", error)
   }
 }
+
+export const tree = async () => {
```
damassi, line 243: this can be moved to `sync/utils/tree`

damassi: (Or even deleted, no need for two functions where one only shadows the other)

src/app/system/sync/fileCache.ts
```diff
@@ -212,11 +231,48 @@ export const loadUrlMap = async () => {
...
+  return await makeTree(DocumentDirectoryPath)
+}
+
+export type TreeFile = { name: string; size: number; prettySize: string }
+export type TreeDirectory = TreeFile & { children: Record<string, TreeNode> }
+export type TreeNode = TreeFile | TreeDirectory
+
+const makeTree = async (path: string) => {
```
damassi, line 251: this can be moved to sync/utils/tree.ts

src/deepPartial.ts
```diff
@@ -0,0 +1,7 @@
+declare module "" {
+  global {
+    type DeepPartial<T extends object> = Partial<{
+      [k in keyof T]: T[k] extends object ? DeepPartial<T[k]> : T[k]
+    }>
+  }
+}
```
damassi, line 7: Can these typings be moved to a `src/typings` folder like we do in Force? https://github.com/artsy/force/tree/main/src/Typings (vs cluttering up the src root)

damassi (approved):
Looks great @pvinis - only one more general comment about avoiding encoding implementation details and one comment about creating a new `tree` util file but other than that 🚢

src/app/components/AddItemsToAlbum.tsx
damassi: (Me just thinking this through, nothing to change here)

pvinis: oh yea, for sure! this will be all hidden once we move this to palette-mobile. the "API" is still I guess not final, which is why i haven't bothered to hide the implementation details. it will def get hidden and simplified once Screen moves to palette-mobile 👍 (which will be soon. i just need a bit of time to add some nice storybook things and we are done.)

src/app/Providers.tsx
damassi: This is a funny provider, i wonder what its doing under the hood 👀

src/app/components/AddItemsToAlbum.tsx
damassi: nice yeah, that's a good point that once its behind a lib name it will all be importable from the package name root 👍

PR #397, opened 2022-12-22
refactor(nav): Simplify navigation

damassi:
### Description

This simplifies the navigation a bit by:
- Hoisting the `ErrorBoundaryWrapper` up a level to just above the nav. Errors thrown from children should bubble up per react design, eliminating the need to wrap each individual route in its own error boundary. (I could be missing something here, but in my tests errors worked just as expected)
- Removing the somewhat unusual `children={() => ...}` form in the `Screen` component in favor of `component={Foo}` 
- Moving all developer tool storybook routes into their own `StorybookNavigation` file

cc @artsy/mobile-platform 

### PR Checklist

### To the reviewers 👀

[docs]: /docs/README.md


.husky/pre-commit
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env sh
 . "$(dirname -- "$0")/_/husky.sh"
 
-npx lint-staged
+npx lint-staged --no-stash
```
damassi, line 4: If you `ctrl+c` while lint staged is running and before the git lockfile has been deleted it is super possible to lose work. Happened to me probably > 10 times, and just happened in this PR. 

With `--no-stash` one can ctrl+c at any time and it just... leaves things alone

src/app/Navigation.tsx
```diff
@@ -118,104 +97,45 @@ export const Main = () => {
...
-            <Screen
-              name="StorybookScreenFullWidthItem"
-              children={() => ErrorBoundaryWrapper(StorybookScreenFullWidthItem)}
-            />
-            <Screen
-              name="StorybookScreenRawHeader"
-              children={() => ErrorBoundaryWrapper(StorybookScreenRawHeader)}
-            />
+            {/* Dev */}
+            {StorybookNavigation()}
```
damassi, line 135: Its annoying that react-navigation is so strict about component wrappers. Hack around it like this.

src/app/App.tsx
```diff
@@ -2,13 +2,16 @@ import { LogBox } from "react-native"
...
+import { ErrorBoundaryWrapper } from "app/system/wrappers/ErrorBoundary"
 
 setupFlipper()
 
 LogBox.ignoreLogs(["lineHeight", "borderRadius", "fontSize", "onBackground", "borderWidth"])
 
 export const App = () => (
   <AppProviders>
-    <Main />
+    <ErrorBoundaryWrapper>
```
damassi, line 13: Unless i'm misunderstanding something about Energy / React Native, this will catch everything in the same layered sort of way as Suspense boundaries catch things; errors will bubble up until the closest boundary. So no need to wrap every individual route in `Navigation.tsx`. 

Please let me know if i'm doing something dumb here!

src/app/App.tsx
```diff
@@ -2,13 +2,16 @@ import { LogBox } from "react-native"
...
 setupFlipper()
 
 LogBox.ignoreLogs(["lineHeight", "borderRadius", "fontSize", "onBackground", "borderWidth"])
 
 export const App = () => (
   <AppProviders>
-    <Main />
+    <ErrorBoundary>
+      <Main />
+    </ErrorBoundary>
```
damassi, line 15: Unless I'm misunderstanding something about Energy / React Native, this will catch everything in the same layered sort of way as Suspense boundaries catch things; errors will bubble up until the closest boundary. So no need to wrap every individual route in `Navigation.tsx`. 

Please let me know if i'm doing something dumb here!

src/app/screens/HomeTabs/Settings/Settings.tsx
```diff
@@ -22,7 +23,7 @@ import { Screen } from "palette"
 
 export const Settings = () => {
   const { relayEnvironment } = useSystemRelayEnvironment()
-  const navigation = useNavigation<NavigationProp<NavigationScreens>>()
+  const navigation = useNavigation<NavigationProp<NavigationScreens & StorybookScreens>>()
```
pvinis, line 26: heh nice

src/app/Navigation.tsx
```diff
@@ -118,104 +97,45 @@ export const Main = () => {
...
           </>
         )}
       </Navigator>
     </NavigationContainer>
   )
 }
-
-const ErrorBoundaryWrapper = (Component: FC) => (
```
pvinis, line 217: oooo youre talking about this one. these are fine to go. i meant the error around the ArtistArtworks for example.

src/app/Navigation.tsx
```diff
@@ -35,33 +27,20 @@ import { ShowTabs } from "app/screens/HomeTabs/Shows/ShowTabs/ShowTabs"
 import { useNetworkStatusListener } from "app/system/hooks/useNetworkStatusListener"
 import { GlobalStore } from "app/system/store/GlobalStore"
 import { loadUrlMap } from "app/system/sync/fileCache"
-import { ErrorBoundary } from "app/system/wrappers/ErrorBoundary"
 
-export type PreAuthScreens = {
+export type AuthScreens = {
```
pvinis, line 31: ok yea that makse sense. i thought like "pre-authed screens" and "authed screens", but yea, better this way.

src/app/Navigation.tsx
```diff
@@ -35,33 +27,20 @@ import { ShowTabs } from "app/screens/HomeTabs/Shows/ShowTabs/ShowTabs"
...
-  DarkModeSettings: undefined
-  EditPresentationMode: undefined
-  ArtistTabs: { slug: string; name: string }
-  AlbumTabs: { albumId: string }
-  InstallImage: { url: string }
-  Artwork: { slug: string; contextArtworkSlugs?: string[] }
-  Search: undefined
-  AlbumArtworks: { albumId: string }
-  ShowTabs: { slug: string }
+export type Screens = {
```
pvinis, line 35: maybe this though?
```suggestion
export type AuthedScreens = {
```

pvinis (changes requested):
lets move the ErrorBoundary inside the Providers.tsx, above the suspensewrapper, and we good!

src/app/Navigation.tsx
damassi: Seems like an unnecessary extra word (I notice that a lot throughout the code, just as a general bit of feedback), since `AuthScreens` implies a gate

damassi:
Updated @pvinis 👍

src/app/App.tsx
damassi: Yup, this makes sense, we can fine-tune this a bit to lean into nesting a bit vs wrapping every route (which seems a bit premature right now). 

Moved the error boundary up to the provider tho, which does make sense to co-locate everything there.

PR #409, opened 2022-12-23
feat: DevMenu also using react-navigation

pvinis:
This PR resolves [MOPLAT-]

### Description

Ive been meaning to do this for a long time. Happy this is easy now.

### PR Checklist

### To the reviewers 👀

[docs]: /docs/README.md


src/app/Navigation.tsx
```diff
@@ -73,18 +73,18 @@ export const Main = () => {
...
               {SearchNavigation()}
               {SettingsNavigation()}
               {ShowsNavigation()}
-              {DevNavigation()}
-
-              <StackNav.Screen name="HomeTabs" component={HomeTabs} />
             </>
           )}
+
+          {DevNavigation()}
```
damassi, line 87: 👍

src/app/screens/Auth/Login.tsx
```diff
@@ -32,6 +33,8 @@ export const LoginScreen = () => {
   const color = useColor()
   const insets = useSafeAreaInsets()
 
+  useSetupRageShake()
```
damassi, line 36: 😄

src/app/screens/Albums/navigation.tsx
```diff
@@ -23,7 +23,7 @@ export type AlbumNavigationScreens = {
 
 export const AlbumsNavigation = () => {
   return (
-    <>
+    <StackNav.Group>
```
damassi, line 26: Nice https://reactnavigation.org/docs/group 👀

src/app/screens/Auth/Login.tsx
damassi: Thoughts on moving this to the same spot we setup `flipper` (App.tsx)? And can then remove second setup call below

src/app/screens/Dev/navigation.tsx
```diff
@@ -26,28 +28,34 @@ export type DevNavigationScreens = {
 export const DevNavigation = () => {
   return (
     <>
```
damassi, line 30: Can we remove this outer fragment now that we're wrapping in `Group`?

src/app/screens/Dev/useSetupRageShake.ts
```diff
@@ -0,0 +1,29 @@
+import { NavigationProp, useNavigation } from "@react-navigation/native"
+import { useEffect } from "react"
+import { NativeModules } from "react-native"
+import RNShake from "react-native-shake"
+import { NavigationScreens } from "app/Navigation"
+import { GlobalStore } from "app/system/store/GlobalStore"
+
+export const useSetupRageShake = () => {
```
damassi, line 8: We should move this to `app/system/devTools`

damassi (approved):
Couple minor requests but nothing blocking 👍

src/app/screens/Dev/navigation.tsx
pvinis: its two groups here, so i think a fragment makes sense?

pvinis: or you mean to use a group in this line too?

src/app/screens/Dev/useSetupRageShake.ts
pvinis: cool

src/app/screens/Dev/navigation.tsx
damassi: ohh i didn't notice the other one, np 👍

src/app/screens/Auth/Login.tsx
pvinis: we cant, because it needs the useNavigation hook. So basically this would normally go only at the initial screen, but since we have 2 possible entry points, we use it in those two places. login and hometabs.
makes sense?

damassi: yep yep, right, its a hook

PR #417, opened 2022-12-28
feat(skeleton): Add skeleton loading states

damassi:
This PR resolves [MOPLAT-565]

### Description

This PR adds a few skeleton loading components and a pattern for use, which mirrors how we do things over on Palette web:

```tsx
<Skeleton>
  <SkeletonText>Hi</SkeletonText>
  <SkeletonBox width={200} />
</Skeleton> 
```

Additionally, wires up the Animated header component to use said state and creates a skeleton component state for the `ArtistTabs` component. 

Follow-up PR will update other areas with relevant skeleton loading states. 

Also switched the default relay network mode to `store-or-network`, which significantly cleans up the UX in the app, since relay nodes for data can be shared across screens directly from the cache. Example: We fetch a list of artist names when we render the main artist list. Since that list is already there, relay will automatically use those list items when rendering the artist detail view. 

 **Detail of fade animation**

https://user-images.githubusercontent.com/236943/209857407-56586cf9-a4d6-4b2b-82b0-4211482c69e6.mp4

**Screens in action**

https://user-images.githubusercontent.com/236943/209857430-3c28a65b-bbd3-489c-afdd-15bb86f74f9c.mp4

cc @artsy/mobile-platform 

### PR Checklist

### To the reviewers 👀

[docs]: /docs/README.md

[MOPLAT-565]: https://artsyproduct.atlassian.net/browse/MOPLAT-565?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/screens/Artists/ArtistTabs/ArtistTabs.tsx
```diff
@@ -247,3 +249,34 @@ export const artistTabsQuery = graphql`
...
+  const selectModeConfig = useHeaderSelectModeConfig()
+  const isOnline = GlobalStore.useAppState((state) => state.networkStatus.isOnline)
+
+  return (
+    <Screen>
+      <Screen.AnimatedTitleHeader
+        title="isLoading"
+        selectModeConfig={selectModeConfig}
+        hideRightElements={!isOnline}
+        isLoading
```
pvinis, line 263: hm I'd we are doing it that way, why not a new comp like Loading header or something? then we don't need to have the isLoading prop for regular use.

src/palette/organisms/Screen/exposed/LargeTitle.tsx
```diff
@@ -1,16 +1,34 @@
...
 export const LargeTitle = () => {
   const title = useAnimatedHeaderTitle()
 
+  if (!title) {
+    return <PlaceholderLargeTitle />
+  }
+
   return (
-    <Flex height={LARGE_TITLE_HEIGHT} pl="2" justifyContent="center">
+    <Flex height={LARGE_TITLE_HEIGHT} pl="2" justifyContent="center" alignSelf="flex-start">
```
pvinis, line 16: what is this for?

src/app/screens/Artists/ArtistTabs/ArtistTabs.tsx
damassi: removed

src/app/screens/Artists/ArtistTabs/ArtistTabs.tsx
```diff
@@ -247,3 +250,33 @@ export const artistTabsQuery = graphql`
...
 `
+
+export const SkeletonArtistTabs = () => {
+  const selectModeConfig = useHeaderSelectModeConfig()
+  const isOnline = GlobalStore.useAppState((state) => state.networkStatus.isOnline)
+
+  return (
+    <Screen>
+      <Screen.AnimatedTitleHeader
+        title=""
```
pvinis, line 261: I'd say let's use a better value for that. undefined would be my pref. title of "" could be that it's just an empty title, which could happen.

damassi: Follow-up from slack convo, as discussed since we're not really doing anything with empty header titles right now in Folio we can come to this case when we actually need it, and update the code 👍

src/palette/organisms/Screen/exposed/LargeTitle.tsx
damassi: This makes it so that the text-field will map to the width of the text, vs by default filling the screen width

damassi: So the skeleton text width = width of text

damassi:
Going to merge and finish up the rest of the loading screens! Fill follow-up if there are any other comments 👍

PR #421, opened 2022-12-29
chore(typography): Update xs -> sm in various sections of the app

damassi:
**Update:**
- On hold til Jimmy / Design team have a chance to think on this. 

### Description

This goes through and updates a handful of places with `xs` -> `sm` so that the app doesn't feel unnaturally small and somewhat unusable text-size-wise.

This is also a general design cleanup pass, and added in some fade-in animations for when images finish loading to make things feel smoother.

_TODO_: 
- We've got a few tablet/phone checks to return `xs` / `sm` variants and with this change I'm now using `sm` in all the areas and thus ignoring the checks -- so, not sure if it should bump up to `md` on tablet. We can think about this later if folks want to proceed. 

**Before:**

https://user-images.githubusercontent.com/236943/210006273-ba4566ed-4177-46d4-8310-be703ecdb902.mp4

**After:**

https://user-images.githubusercontent.com/236943/210040334-7a9c51ab-3fcf-463d-9820-92e7584a6771.mp4


src/app/system/constants.tsx
```diff
@@ -0,0 +1,4 @@
+/**
+ * Touch area box for UI elements
+ */
+export const DEFAULT_HIT_SLOP = { top: 20, bottom: 20, left: 20, right: 20 }
```
damassi, line 4: At the previous 10px, I was consistently failing to trigger interactions on my phone

pvinis: I wonder if this could be in the Touchable of palette. or if that would mess up things 🤔

src/setupJest.ts
```diff
@@ -165,6 +165,10 @@ jest.mock("react-native-gesture-handler", () => {
   }
 })
 
+jest.mock("app/system/wrappers/CachedImage", () => ({
+  CachedImage: jest.requireActual("react-native").Image,
```
pvinis, line 169: nice

src/app/system/constants.tsx
damassi: For a mobile app 20px seems like the minimum! If we have a lot of `hitSlop` definitions in there then having a constant that's referenced and exported could make sense. (On palette web for example we export things like drop shadows... that kinda thing)

src/app/system/wrappers/CachedImage.tsx
```diff
@@ -1,22 +1,63 @@
...
+  // needed to prevent the opacity from resetting to 0 on rerenders.
+  let ImageWrapper: any
+  let styleProps
+  if (isDoneLoading.current === true) {
+    ImageWrapper = Image
+    styleProps = [style]
+  } else {
+    ImageWrapper = Animated.Image
+    styleProps = [style, animatedStyle]
+  }
```
damassi, line 52: @pvinis helppp. This was pretty maddening. Do you know a better way to do this? This doesn't feel terrible, but yah `useSharedValue` would always get reset to its initial value on re-render. (For example, when in Image screen and sliding left to the next image, opacity would reset back to zero). I ran into this issue: https://github.com/software-mansion/react-native-reanimated/issues/3224

src/app/screens/Artwork/ArtworkContent/ArtworkContent.tsx
```diff
@@ -66,7 +79,7 @@ export const ArtworkContent = ({ slug }: { slug: string }) => {
   const imageFlexHeight = screenHeight - BOTTOM_SHEET_HEIGHT - NAVBAR_HEIGHT
 
   // eslint-disable-next-line react-hooks/exhaustive-deps
-  const snapPoints = useMemo(() => [BOTTOM_SHEET_HEIGHT, "95%"], [screenHeight])
+  const snapPoints = useMemo(() => [BOTTOM_SHEET_HEIGHT, "94%"], [screenHeight])
```
damassi, line 82: Fixes issue where bottom panel would slide up and kinda cover back button circle / be too close to top of screen

pvinis (approved):

PR #428, opened 2023-01-03
chore(polish): Add UI polish

damassi:
> ⚠️ Depends on https://github.com/artsy/palette-mobile/pull/33

### Description

This adds some UI polish/cleanup to the app, including slightly refactoring the bottom artwork bottom sheet to be less brittle. Also removes the Skeleton component in favor of the one from `palette-mobile` and adds a subtle fade-in when images are done loading. 

cc @artsy/mobile-platform 

### PR Checklist

### To the reviewers 👀

[docs]: /docs/README.md


package.json
```diff
@@ -33,7 +33,7 @@
     "react-native-pager-view": "Peer dep of react-native-collapsible-tab-view@5"
   },
   "dependencies": {
-    "@artsy/palette-mobile": "3.4.0",
+    "@artsy/palette-mobile": "3.4.1--canary.33.3833203630.0",
```
damassi, line 36: ⚠️ Shouldn't merge until this is updated!

src/app/components/ImageModal.tsx
```diff
@@ -29,9 +29,10 @@ export const ImageModal = ({ isModalVisible, setIsModalVisible, uri }: ImageModa
             style={{ flex: 1, width: "100%" }}
             resizeMode="contain"
             placeholderHeight={undefined}
+            fadeInOnLoad={false}
           />
         </ReactNativeZoomableView>
-        <Flex position="absolute" pt={insets.top} px={2}>
+        <Flex position="absolute" pt={insets.top + 8} left={-8} px={2}>
```
pvinis, line 35: where are these numbers coming from?

src/app/system/wrappers/CachedImage.tsx
```diff
@@ -1,22 +1,62 @@
...
 }
 
 export const CachedImage: React.FC<CachedImageProps> = ({
+  fadeInOnLoad = true,
   placeholderHeight,
+  style,
   uri,
   ...restProps
 }) => {
+  const isDoneLoading = useRef(false)
```
pvinis, line 25: i guess this could also be `useSharedValue`, for consistency

src/app/system/wrappers/CachedImage.tsx
```diff
@@ -1,22 +1,62 @@
...
+  const opacity = useSharedValue(fadeInOnLoad ? 0 : 1)
+  const animatedStyle = useAnimatedStyle(() => ({ opacity: opacity.value }), [])
```
pvinis, line 27:
```suggestion
  const fadeInAnim = useAnimatedStyle(() => ({ opacity: opacity.value }), [])
```
this is a bit nicer as a convention

src/app/components/ImageModal.tsx
damassi: oh i wonder if this was carry-over from the other PR when i was fixing the sizes, i can remove

pvinis: thanks

src/app/system/wrappers/CachedImage.tsx
damassi: This is mentioned below, but unfortunately we need an additional piece of state here since useSharedValue will reset state on rerender and set the image opacity to zero 😢 https://github.com/software-mansion/react-native-reanimated/issues/3224. Having this as an intermediate fixes it. Can see the problem clearly when swiping between artworks on artwork detail view

src/app/components/ImageModal.tsx
damassi: Oh, i know what this is! This is so that the back (close) button in the image modal lines up exactly with the back button on the view. Without this there's a shift (though I don't like the magic numbers)

package.json
pvinis: i think this will end up being 3.5.2

src/app/components/ImageModal.tsx
pvinis: hm. i wonder if we can fix this differently. but if you don't wanna do this now, we can skip it and figure it out when it breaks next time :D

src/app/system/wrappers/CachedImage.tsx
pvinis: aaaah, that was the comment about. cool cool.

pvinis (approved):

pvinis:
assigning back to you. merge when ready

PR #431, opened 2023-01-04
feat: with RetryErrorBoundary and fetchKeys

rajsam003:
This PR resolves [MOPLAT-541]

This PR adds the retry button in the error screen. The implementation is referred from https://relay.dev/docs/guided-tour/rendering/error-states/#when-using-uselazyloadquery

- Using retry method in the ErrorBoundary file 
- Renaming ErrorBoundary file to RetryErrorBoundary
- Adding `<GlobalRetryErrorBoundary>` as parent provider of all providers.
- Introducing `<TabsScreenWrapper>` for Tab's screen which has `<RetryErrorBoundary>` and `<SuspenseWrapper>` 
- Replaced `useLazyLoadQuery()` with `useSystemQueryLoader()`
- Adding `relayFetchKey` in NetworkStatusModel

image

### PR Checklist

- [x] I added Tests and Stories for my changes.

### To the reviewers 👀

[docs]: /docs/README.md

[MOPLAT-541]: https://artsyproduct.atlassian.net/browse/MOPLAT-541?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/components/GlobalErrorView.tsx
```diff
@@ -0,0 +1,13 @@
+import { Text, View } from "react-native"
+
+interface GlobalErrorViewProps {
+  error: Error
+}
+
+export const GlobalErrorView = ({ error }: GlobalErrorViewProps) => {
+  return (
+    <View style={{ flex: 1, justifyContent: "center", alignItems: "center" }}>
+      <Text>Error: {error?.message}</Text>
```
damassi, line 10: Ditto about palette

src/app/components/Items/AlbumListImage.tsx
```diff
@@ -8,11 +8,16 @@ import { imageSize } from "app/utils/imageSize"
 interface AlbumListImageProps {
   slug: string
   style?: ImageProps["style"]
+  fetchKey: number
 }
 
-export const AlbumListImage = ({ slug, style }: AlbumListImageProps) => {
+export const AlbumListImage = ({ slug, style, fetchKey }: AlbumListImageProps) => {
```
damassi, line 14: As mentioned in slack, we shouldn't have to pass fetch key's all around the app. They can be stored in global state and then managed within the `useSystemQueryLoader` (with these `useLazyLoadQueries` being replaced.

src/app/components/LoadFailureView.tsx
```diff
@@ -0,0 +1,81 @@
...
+import { debounce } from "lodash"
+import { useRef, useState } from "react"
+import { Animated, Easing } from "react-native"
+
+interface LoadFailureViewProps {
+  error?: Error
+  onRetry?: () => void
+}
+
+export const LoadFailureView: React.FC<LoadFailureViewProps & BoxProps> = ({
```
damassi, line 11: Since this is ultimately tied to an `ErrorBoundary`, we should keep the language the same: `LoadErrorView` (vs mixing "failure" and "error").

damassi (changes requested):
I think we can make this approach simpler and less brittle! Happy to work through the implementation together and brainstorm.

In general, it's the sort of thing where anytime we're doing something numerous times its good to step back and ask: can we automate this? Can we make it transparent so that future devs don't need to add additional wiring to create new views? Because every time a new API addition like this gets added, it creates another thing that we need to worry about and refactor should something change underneath.

src/app/components/GlobalErrorView.tsx
```diff
@@ -0,0 +1,13 @@
+import { Text, View } from "react-native"
+
+interface GlobalErrorViewProps {
+  error: Error
+}
+
+export const GlobalErrorView = ({ error }: GlobalErrorViewProps) => {
+  return (
+    <View style={{ flex: 1, justifyContent: "center", alignItems: "center" }}>
```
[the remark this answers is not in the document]
rajsam003: We have wrapped all the providers with `<GlobalRetryErrorBoundary>`. ThemeProvider is one of them, which provides theme for palette component `<Text>`.
As `<GlobalErrorView>` is the child component of `<GlobalRetryErrorBoundary>`. We cannot use `<Text>` from palette in it.  I've used `<Text>` from `react-native` lib. Also <Flex>. Just to import both from the same library.

src/app/components/GlobalErrorView.tsx
rajsam003: addressed the reason in the above comment.

src/app/components/Items/AlbumListImage.tsx
rajsam003: Good idea. Can we share the same global fetchKey to all the queries? Just wondering.

rajsam003:
> I think we can make this approach simpler and less brittle! Happy to work through the implementation together and brainstorm. See [this slack thread](https://artsy.slack.com/archives/C02NAJ2CGLW/p1672857367275759) for a possible approach.
> 
> In general, it's the sort of thing where anytime we're doing something numerous times its good to step back and ask: can we automate this? Can we make it transparent so that future devs don't need to add additional wiring to create new views? Because every time a new API addition like this gets added, it creates another thing that we need to worry about and refactor should something change underneath.

Thanks @damassi :). I totally agree. We should always try to automate code to make it easier for future devs.

src/app/components/GlobalErrorView.tsx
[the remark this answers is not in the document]
damassi: @rajsam003 - any chance this error boundary can go _after_ `<Theme>`, or move `<Theme>` to the top?

src/app/components/GlobalErrorView.tsx
damassi: https://github.com/artsy/energy/pull/431#discussion_r1062684783

src/app/components/Tabs/TabsScreenWrapper.tsx
```diff
@@ -0,0 +1,10 @@
+import { RetryErrorBoundary } from "app/system/wrappers/RetryErrorBoundary"
+import { SuspenseWrapper } from "app/system/wrappers/SuspenseWrapper"
+
+export const TabsScreenWrapper: React.FC = ({ children }) => {
```
damassi, line 4: We can probably simplify the name! No need to add `Wrapper`. 

```
<TabScreen>
  ...
</TabScreen>
```

feels nice and clean

damassi (approved):
Looks great @rajsam003 -- Thanks for working through those updates 👍 

One suggestion about updating the naming of that one component to just `<TabScreen>` but otherwise LGTM

src/app/components/GlobalErrorView.tsx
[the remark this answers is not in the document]
pvinis: it could

PR #433, opened 2023-01-04
refactor(tests): Refactor tests to use artsy setupTestWrapper pattern

damassi:
### Description

This updates Energy's tests with the `setupTestWrapper` relay test pattern we see around various artsy apps, which simplifies how we do testing overall. Also updated tests to be more "real world", in the sense that we no longer have two app provider wrappers executing -- one for tests, and one for real app -- and instead use one and share them between.

cc @artsy/mobile-platform 

### PR Checklist

### To the reviewers 👀

[docs]: /docs/README.md


src/app/Providers.tsx
```diff
@@ -3,19 +3,24 @@ import JotaiNexus from "jotai-nexus"
...
+}
+
+export const AppProviders: React.FC<AppProvidersProps> = ({ children, relayEnvironment }) => {
   return (
     <ErrorBoundary>
       <GlobalStoreProvider>
         <ThemeProvider>
           <SuspenseWrapper>
-            <RelayProvider>
+            <RelayProvider relayEnvironment={relayEnvironment}>
```
damassi, line 23: Since we're now sharing the app provider component across tests, we need to be able to pass a test environment in.

src/app/system/devTools/ignoreLogs.ts
```diff
@@ -0,0 +1,45 @@
...
+const IGNORED_LOGS = [
+  "lineHeight",
+  "borderRadius",
+  "fontSize",
+  "onBackground",
+  "borderWidth",
+  "react-native-flipper",
+]
+
+const IGNORED_ERRORS = ["inside a test was not wrapped in act(...)"]
```
damassi, line 12: This is a consistent red-herring error and ignoring it is harmless

src/app/utils/test/mockEnvironmentPayload.ts
```diff
@@ -46,23 +83,3 @@ const defaultMockResolvers: MockResolvers = {
...
-const createTestEnvironment = () => {
-  const mockEnvironment = createMockEnvironment()
-  return mockEnvironment
-}
-
-export let relayMockEnvironment = createMockEnvironment()
-
-export const resetRelayMockEnvironment = () => {
-  relayMockEnvironment = createTestEnvironment()
-}
```
damassi, line 59: We don't need this setup scaffolding anymore since we're now creating the environment when setupTestWrapper is initially invoked

src/setupJest.ts
```diff
@@ -1,14 +1,17 @@
-// MARK: - General preparation
-
 import "@testing-library/jest-native/extend-expect"
+import mockAsyncStorage from "@react-native-async-storage/async-storage/jest/async-storage-mock"
+import React from "react"
+import { ignoreLogsInTests } from "app/system/devTools/ignoreLogs"
+import { ScreenDimensionsWithSafeAreas } from "app/utils/hooks/useScreenDimensions"
```
damassi, line 5: Having imports spread throughout a file is confusing with VSCode auto-importer and is also invalid JS (ignoring babel of course)

src/app/utils/test/mockEnvironmentPayload.ts
```diff
@@ -1,10 +1,47 @@
...
+    if (preloaded) {
+      if (!query) {
+        throw new Error("A `query` is required when using `preloaded` prop.")
+      }
+      mockEnvironment.mock.queuePendingOperation(query, variables)
+    }
+
+    resolve((operation: any) => {
+      return MockPayloadGenerator.generate(operation, {
+        ...defaultMockResolvers,
```
damassi, line 34: I need to move this whole file over to force!

MounirDhahri: good idea 😄

src/app/utils/test/setupTestWrapper.tsx
```diff
@@ -0,0 +1,120 @@
...
+  preloaded?: boolean
+  query?: GraphQLTaggedNode
+  variables?: T["variables"]
+}
+
+/**
+ * Creates a test renderer which can be used to render a variety of relay
+ * query configurations.
+ *
+ * @example - Full Queries
```
MounirDhahri, line 20: **praise:** thanks for the examples!

src/app/utils/test/setupTestWrapper.tsx
```diff
@@ -0,0 +1,120 @@
...
+      return (
+        <>
+          {query ? (
+            <QueryRenderer<T>
+              environment={mockEnvironment}
+              variables={variables}
+              query={query}
+              render={({ props, error }) => {
+                if (props) {
+                  return <Component {...props} />
```
MounirDhahri, line 93: **nitpick:** might be worth it at some point to also support injecting some initial props here

MounirDhahri (approved):
LGTM! thanks Chriss for this nice refactoring

src/app/components/Items/ArtworkGridItem.tsx
```diff
@@ -77,7 +77,7 @@ export const ArtworkGridItem = ({
   )
 }
 
-const ArtworkGridItemFragment = graphql`
+export const ArtworkGridItemFragment = graphql`
```
gkartalis, line 80: Do we need to export this? Don't see it being used anywhere 🤔

src/app/utils/test/setupTestWrapper.tsx
```diff
@@ -0,0 +1,120 @@
...
+ * })
+ *
+ * it('works', () => {
+ *   const { getByText } = await renderWithRelay()
+ *   expect(getByText('name')).toBeTruthy()
+ * })
+ *
+ * @example - Using fragments
+ *
+ * const Bar = () => {
```
gkartalis, line 45: **question(non-blocking):** What about the cases like [this](https://github.com/artsy/eigen/blob/7386017bd82484524898ec0e76a8b0b0fde9c38f/src/app/Scenes/MyProfile/MyProfileHeader.tsx#L24) (picked up a random one from eigen) that we are passing the `$key`? 

Asking because I tried to refactor some tests in eigen the other day and not passing the `query` but they didn't work 🤔

gkartalis (approved):

rajsam003 (approved):
Looks good to me. Thanks for refactoring 🙏🏽

src/app/utils/test/setupTestWrapper.tsx
damassi: Its not possible to test fragments in this way! Need a query for fragments unless you're importing a query from somewhere that includes the fragment under test.  (That's why I added this second example in the docs). Testing lazy queries tho you don't need the fragment. 

Worth noting too that I modified `setupTestWrapper` slightly in Energy from where its seen elsewhere to make it `async`. In a few places in the past I noticed that we'd need to use `waitFor` in our test, but now that its async (with `flushPromiseQueue` we no longer need to do this. So i'd give a look at this version of the `setupTestWrapper` file and see if the changes might help in some way.

gkartalis: Thanks!

MounirDhahri:
Sorry @damassi I didn't notice I was the assignee for the PR 🙌

PR #435, opened 2023-01-06
test(offline): Add test coverage

damassi:
This PR resolves [MOPLAT-548]

### Description

This adds test coverage to the data portion of the app! Looking at the syncManager code, the only bits we need to test are the parsers. Everything else is sufficiently covered by Relay’s auto-generated types and is in effect logic-less — meaning: if ever anything were to change anywhere with our queries, be it name or variables passed or data returned, it would instantly blow up the type-checker in numerous places, including the syncManager. The contract in the component code is sufficient to cover the contract in the offline sync code. With the parsers however we do transform data, and that is subject to breakage.

(Note this only covers the data portion off offline support, not any of the image sync behaviors.) 

cc @artsy/mobile-platform 

### PR Checklist

### To the reviewers 👀

[docs]: /docs/README.md

[MOPLAT-548]: https://artsyproduct.atlassian.net/browse/MOPLAT-548?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/system/sync/syncManager.ts
```diff
@@ -547,3 +539,8 @@ export const loadRelayDataFromOfflineCache = (
     resetRelayEnvironment(JSON.parse(data!))
   })
 }
+
+export const tests = {
+  syncResults,
+  parsers,
```
damassi, line 545: lets keep internal functions private

src/app/system/sync/syncManager.ts
```diff
@@ -547,3 +539,9 @@ export const loadRelayDataFromOfflineCache = (
     resetRelayEnvironment(JSON.parse(data!))
   })
 }
+
+export const tests = {
```
pvinis, line 543: there are a few places we do this https://github.com/artsy/eigen/blob/main/src/palette/Theme.tsx#L325-L329. maybe this naming could work here too? obviously non-blocking.

rajsam003 (approved):

src/app/system/sync/syncManager.ts
damassi: oh nice, I like this pattern! Seeing `export const foo` everywhere can often confuse intent but wrapping privates within a tests export makes it nice and clear

damassi: Will update naming since there's already an established pattern 👍

PR #443, opened 2023-01-10
feat: with RetryErrorBoundary and fetchKeys

damassi:
Related artsy/energy#442.

Re-opens PR so that fixes can be looked at!


rajsam003:
Thanks for this bringing back this PR again <3

src/app/screens/Artwork/Artwork.tsx
```diff
@@ -47,15 +48,9 @@ export const Artwork = () => {
 
   const screens: ScrollableScreenEntity[] = artworkSlugs.map((slug) => ({
     name: slug,
-    content: (
-      <ErrorBoundary withoutBackButton>
-        <SuspenseWrapper key={slug}>
```
rajsam003, line 52: This was causing the issue. I dont think we need it.

PR #445, opened 2023-01-10
fix(offline): Fix images failing to return sporadically

damassi:
### Description

Silly bug: `&&` -> `||`. This fixes an issue that we were seeing around images periodically failing to load when offline. It was due to a race condition, in that before in image was ready to load by retrieving the path from the offline cache the image would render (though with an invalid url) and then immediately update. By catching that render and showing a fallback first, we're able to successfully load the image once its been retrieved from the file system. 

One thing this _has_ revealed to us however is that the offline image sync itself can be flimsy -- meaning, sometimes the image requests 401 out and are never saved, even after a retry. Not sure why, since some work just fine? Maybe it has something to do with Gemini rate limiting our request given that we're throwing a ton of requests at it in parallel. I'll ping Velocity to follow up on this and continue investigating in a follow-up PR. 

This PR also does a few other things:
- Adds retry support to file downloads
- Makes ID lookups a little safer by filtering for bad values 
- Adds an `useIsOnline` hook since we're using that bit of code all over at this point

cc @artsy/mobile-platform 

### PR Checklist

### To the reviewers 👀

[docs]: /docs/README.md


src/app/system/wrappers/CachedImage.tsx
```diff
@@ -31,7 +31,7 @@ export const CachedImage: React.FC<CachedImageProps> = ({
     opacity.value = withTiming(1, { duration: 200, easing: Easing.ease })
   }
 
-  if (!uri && !localUri) {
+  if (!uri || !localUri) {
```
pvinis, line 34: all the more reason for not using these `!whatever` forms. looking at this I am still not sure whats up, whereas something like `if (uri === undefined || localUri === undefined) {` makes it SOOOOOO much more readable.

src/app/system/sync/fileCache.ts
```diff
@@ -70,47 +61,72 @@ export const downloadFileToCache = async ({
...
+    try {
+      await makeSureCacheIsReady()
+
+      const id = uuidv4()
+
+      const filename = (() => {
+        switch (type) {
+          case "image": {
+            const src = parse(url).src as string
+            const imageExtension = src?.substring(src.lastIndexOf("."))
```
damassi, line 82: This seems to be working perfect for image stuff, lemme know if I'm missing something @pvinis

src/app/system/sync/syncManager.ts
```diff
@@ -342,38 +344,38 @@ export function initSyncManager({
 
     const urls = parsers.getImageUrls()
 
-    await forEachAsync(urls, (url) =>
-      downloadFileToCache({
+    await forEachAsync(urls, (url) => {
+      return downloadFileToCache({
```
damassi, line 348: Was debugging and of course had to unwind these implicit returns to add logs 😅  Its best to assume that code is always, 100% of the time, going to be returned to in some form or another in the future and to just make future developer selves happy by not short-handing stuff. (Assuming the code is multi-line; single-line implicit returns are generally never debugged in practice and make total sense.)

damassi:
@pvinis - gonna merge this but will follow up with any further comments or concerns that you might have when your back online

src/app/system/sync/fileCache.ts
```diff
@@ -70,47 +61,72 @@ export const downloadFileToCache = async ({
...
+          `[fileCache] Error downloading file: ${error}. Retrying [${errorRetryAttempt}]: ${url}`
+        )
 
-    urlMap[url] = filename
-    writeAtom(urlMapAtom, urlMap)
-    await saveUrlMap()
-  } catch (error) {
-    console.log("[fileCache] Error downloading file:", error)
+        // Retry download
+        await startDownload()
```
MounirDhahri, line 124: beautiful 🤌 a retry strategy could be beneficial here

damassi: I just <3 javascript. Sometimes all that one needs to do something seemingly complicated is a second function and a closure.

PR #475, opened 2023-01-26
fix: android crash

brainbicycle:
This PR resolves [MOPLAT-556]

### Description

Hermes doesn't support named capture groups in regexes which is leading to a crash in the app. Just removes it since we only have it in one place. 
https://github.com/facebook/hermes/issues/46

Follow-up: When I build a release variant locally this is a build time error, so there must be some delta with how we are building/bundling in ci which I don't love so I will follow-up to check what is going on here.

### PR Checklist

### To the reviewers 👀

[docs]: /docs/README.md

[MOPLAT-556]: https://artsyproduct.atlassian.net/browse/MOPLAT-556?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pvinis (approved):
nice catch!

PR #483, opened 2023-01-31
fix(folio-release): Fix screen flash and clear cache when logged out

damassi:
This PR resolves [MOPLAT-595]; also fixes [MOPLAT-585]

### Description

This does a few things: 
- Delete offline cache when logged out 
- Fixes the screen flash issue when clicking select all on artist > artworks 

https://user-images.githubusercontent.com/236943/215895167-d3d1d0b9-b663-459e-9d0d-3f48dc14270d.mp4

cc @artsy/mobile-platform 

### PR Checklist

### To the reviewers 👀

[docs]: /docs/README.md

[MOPLAT-595]: https://artsyproduct.atlassian.net/browse/MOPLAT-595?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MOPLAT-585]: https://artsyproduct.atlassian.net/browse/MOPLAT-585?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/palette/organisms/Screen/atoms.ts
```diff
@@ -27,7 +27,11 @@ const useSetterHook = <T>(atom: PrimitiveAtom<PerScreen<T>>, value: T): void =>
   const screenName = useScreenName()
   const [values, setValues] = useAtom(atom)
 
-  useEffect(() => setValues({ ...values, [screenName]: value }), [value])
+  useEffect(() => {
+    if (value) {
```
damassi, line 31: I imagine this will be some future bug, but until then this looks good in the app

src/app/screens/Settings/Settings.tsx
```diff
@@ -65,7 +65,7 @@ export const Settings = () => {
             </Flex>
           </Touchable>
 
-          <Button block onPress={() => void GlobalStore.actions.auth.signOut()}>
```
pvinis, line 68: is this bad? i thought it was good practice.

src/app/system/store/Models/DevicePrefsModel.ts
```diff
@@ -22,16 +23,35 @@ export interface DevicePrefsModel {
 
   lastSync: string | null
   setLastSync: Action<this, DateTime | null>
+
+  clearCacheOnSignOut: ThunkOn<this, null, GlobalStoreModel>
+  offlineSyncedChecksum: string | null
```
pvinis, line 28: the reason i had this as just string was that it felt like an unnecessary check when we know it will always be a string, and i initialize it with a silly one. either way is fine, just mentioning my thought for the previous type.

src/palette/organisms/Screen/atoms.ts
pvinis: whats the issue here? what was happening and what is happening now?

src/app/screens/Settings/OfflineModeSettings.tsx
```diff
@@ -112,25 +114,26 @@ export const OfflineModeSettings = () => {
...
                   Your synced data needs to be refreshed. Please tap the "Start sync" button above.
                 </Text>
               </>
             )}
+
             {(isUserDev || __DEV__) && (
               <>
-                <Spacer y="1" />
-                <Text color="onBackgroundLow">Latest synced: {offlineSyncedChecksum}</Text>
+                <Text color="onBackgroundLow">Last sync: {offlineSyncedChecksum}</Text>
```
pvinis, line 136: "last sync" is the date above, this is the latest synced checksum 🤔. non-blocking obviously.

src/app/screens/Artists/ArtistTabs/ArtistTabs.tsx
```diff
@@ -50,7 +50,9 @@ export const ArtistTabs = () => {
 
   const artworkData = useSystemQueryLoader<ArtistTabsQuery>(artistTabsQuery, {
     partnerID,
-    artworkIDs: selectedWorks,
+    // TODO: Do we really need this for fetching artworks? The flatlist displays
+    // all of our artworks regardless. Revisit if we start doing lazy load.
+    artworkIDs: [], // selectedWorks,
```
pvinis, line 55: is this for the selection flash? what happened there? was this serving any purpose? 🤔

src/app/screens/Settings/Settings.tsx
damassi: Never really see this in react TS codebases 🤷 just removes a little potential future indirection due to the its uncommonness

src/palette/organisms/Screen/atoms.ts
damassi: For some strange reason in a rerender loop (and HMR reset loop) state is being reset to the default state, which was falsy. Right now we're not setting these atoms to null anywhere so seems to be fine

src/app/screens/Settings/Settings.tsx
pvinis: cool. ive definitely had bugs before because of this missing, and i either had to do `() => { whatever() }` or `() => void whatever()`, and i preferred the void way.

src/app/screens/Artists/ArtistTabs/ArtistTabs.tsx
damassi: Yep! this fixed the flash. The way it worked was that it would pass selected ids into the query to narrow down the fetch, but that's actually not necessary since the list is already there and will stay there regardless. By not passing in the ids the query remains consistent (and cached) and the flash goes away.

PR #497, opened 2023-02-03
feat(migrations): Add state migrations

damassi:
This PR resolves [MOPLAT-478]

### Description

This adds state migrations to Folio, which behave identically to how we do it in Eigen. Along the way we found a chance to simplify the implementation however, and rely exclusively on easy-peasy's persistence implementation, versus needing to roll our own as we did in Eigen. 

Migrations live in `app/system/migrations.ts`. 

Still need to do a little cleanup and add some tests/docs, but this is ready for review. 

Follow-up is moving this new implementation back over to Eigen and remove all of the hand-rolled persistence code. 

**To Test**:
- Via flipper, install and enable `flipper-async-storage-advanced`
- Start energy and add a new migration
- Go into the flipper async storage pane and change `version` to something lower, hit update button
- Reload Energy in the simulator 

cc @artsy/mobile-platform 

### PR Checklist

### To the reviewers 👀

[MOPLAT-478]: https://artsyproduct.atlassian.net/browse/MOPLAT-478?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


package.json
```diff
@@ -100,6 +100,7 @@
     "react-spring": "8.0.23",
     "relay-runtime": "^14.1.0",
     "relay-test-utils": "^14.1.0",
+    "rn-flipper-async-storage-advanced": "^1.0.4",
```
damassi, line 103: Gonna remove this

src/app/App.tsx
```diff
@@ -8,6 +9,7 @@ ignoreLogs()
 
 export const App = () => (
   <Providers>
+    <FlipperAsyncStorage />
```
damassi, line 12: And also gonna remove this, just for testing

src/app/Navigation.tsx
```diff
@@ -45,7 +45,7 @@ type Screens = {
 export const StackNav = createStackNavigator<NavigationScreens>()
 
 export const Main = () => {
-  const isRehydrated = useStoreRehydrated()
+  const isDoneBooting = useSystemIsDoneBooting()
```
damassi, line 48: We needed to wrap `easy-peasy`'s hydration hook to ensure that we were done migrating. At that point the app is safe to boot.

src/app/system/store/Models/GlobalStoreModel.ts
```diff
@@ -40,10 +52,22 @@ export const getGlobalStoreModel = (): GlobalStoreModel => ({
...
+    }),
+  },
+
   reset: action((state) => {
     state.auth.activePartnerID = null
   }),
 
+  // Migrates the locally persisted state to the latest version
+  performMigrations: action((state) => {
+    migrateState({ state })
```
damassi, line 68: This is the secret sauce in the PR. Rather than relying on an outside-the-system state / persistence update mechanism operating on JSON objects, just thread it right through easy-peasy's own state update model via an action, which simultaneously updates the persisted async storage in the background.

src/app/system/store/migrations.ts
```diff
@@ -0,0 +1,51 @@
...
+ * IMPORTANT
+ * Before you modify this file please read docs/adding_state_migrations.md
+ */
+
+export const Versions = {
+  DummyMigration: 1,
+  DummyMigration2: 2,
+  DummyMigration3: 3,
+  DummyMigration4: 4,
+  DummyMigration5: 5,
```
damassi, line 14: Will clear this stuff out on Monday but would like you to test / verify things are working good @pvinis 🙏

src/app/system/store/persistence/utils/migrateState.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  while (state.version < toVersion) {
+    const nextVersion = state.version + 1
+    const migrator = migrations[nextVersion]
+
+    if (!migrator) {
+      throw new Error("No migrator found for app version " + nextVersion)
+    }
+
+    migrator(state)
```
damassi, line 26: pass immerized state all the way through to the migrator function; no need for intermediate state representations anymore

src/app/system/store/persistence/migrateState.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  while (state.version < toVersion) {
+    const nextVersion = state.version + 1
+    const migrator = migrations[nextVersion]
+
+    if (!migrator) {
+      throw new Error("No migrator found for app version " + nextVersion)
+    }
+
+    migrator(state)
```
damassi, line 26: Pass immerized state all the way through to the migrator function from the action; no need for intermediate state representations anymore

gkartalis (approved):
Looking great! Only one thing, it would be nice to have a similar doc to the one we have in eigen for [migrations](https://github.com/artsy/eigen/blob/main/docs/adding_state_migrations.md)

src/app/system/store/Models/GlobalStoreModel.ts
MounirDhahri: Nice trick 👏

MounirDhahri (approved):
Nice PR! thanks for assigning me

package.json
pvinis: its not super useful, but i guess its kinda nice to have. though its one of those things that if it is ever bothering/blocking anything, it would go away in an instant.

src/app/App.tsx
pvinis: ah, i remember temporarily adding this like
```
<>
<FlipperThing/>

<Providers>
<Main/>
</Prov>
<>
```
but like.. same. annoying that its a component

src/app/Navigation.tsx
pvinis: what do you think about making a patch to add a `/** @deprecated at Artsy we use useSystemIsDoneBooting(). */` or something? 🤔 so if someone uses it accidentally, they get a strikethrough that might help them figure out the right thing to use.

src/app/system/store/persistence/migrateState.ts
```diff
@@ -0,0 +1,30 @@
+import { CURRENT_APP_VERSION, energyAppMigrations, Migrations } from "app/system/store/migrations"
+
+export function migrateState<State extends { version: number }>({
```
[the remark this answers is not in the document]
pvinis: you mean console log? if yes, then sure

src/app/system/store/Models/GlobalStoreModel.ts
```diff
@@ -20,6 +22,14 @@ interface GlobalStoreStateModel {
   networkStatus: NetworkStatusModel
   presentationMode: PresentationModeModel
   selectMode: SelectModeModel
+
+  // Meta state / actions
+  version: number
+  performMigrations: Action<this>
+  sessionState: {
+    isDonePerformingMigrations: boolean
+    setIsDonePerformingMigrations: Action<GlobalStoreStateModel>
```
pvinis, line 31: where is this called? i cannot find it in this PR 🤔

pvinis (approved):
nice! thanks for taking this over here.

src/app/system/store/persistence/migrateState.ts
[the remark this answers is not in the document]
MounirDhahri: yes, that's what I had in mind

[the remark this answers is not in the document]
pvinis: yep. and actually, it might be a good thing to send to sentry or segment. just a little event maybe 🤔?

pvinis:
oh actually, one thing that is missing here from my branches are the tests. could you add the migration.tests.tsx back here?

src/app/App.tsx
damassi: This is another PR I was thinking about, removing this component requirement. I mean, even if it were a hook or something that'd be nicer than embedding a component.

src/app/Navigation.tsx
damassi: I'm going to really try to pair back patches unless strictly needed, even if they're a comment, across the board. This is something that will only be used once, so not too worried. But even if it weren't, rolling with the `system` concept for apps is something that I've found works really well; if there's a third party lib requirement that interfaces with this level of the stack, roll it into a general "system" init function. Everything goes there, and it becomes a sort of cultural understanding.

src/app/system/store/persistence/migrateState.ts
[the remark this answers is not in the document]
MounirDhahri: +++

damassi:
Yup @pvinis mentioned in the description. Still need to add the tests. Re `isDonePerformingMigrations`, that was something that i refactored late in the day. It will now be assigned in the `performMigrations()` action, no need for a second action for this. Need to update.

package.json
damassi: Moved to dev dep

damassi:
Ok! Addressed feedback and merging this. 

At some point if my state migrations PR to easy peasy is accepted i'll loop back around and strip out all of this unnecessary baggage, and also update Eigen. 🤞 that they merge it.

PR #514, opened 2023-02-14
fix: sentry sourcemaps + bundled js upload

brainbicycle:
This PR resolves [] 

### Description

Fixes uploading js bundles and source maps to sentry. 
- Explicitly calls bundle for both platforms so we have a reliable and easy place the files are output to

Potential follow-ups:
- It is a bit wasteful to bundle again, this happens as part of the build step but is output to different directories depending on Android vs iOS, this is probably configurable somewhere but it is the easy fix for now
- This can be potentially removed from fastlane completely sentry does some automatic uploading of these things now: https://docs.sentry.io/platforms/react-native/manual-setup/manual-setup/#bundle-react-native-code-and-images (one open question before just doing that, we have named releases in sentry when we create betas, would need to figure out how to support that with the new flow)

**Draft while I attempt a beta from this branch** 

### PR Checklist

### To the reviewers 👀


damassi (approved):
Once it's out of draft LGTM 👍

PR #529, opened 2023-02-24
refactor(select-mode): Rewrite select mode functionality using global store

damassi:
### Description

This PR is some prework for a bug that we ran into in the mailer. Noticed that we were using multiple overlapping state implementations, all woven together, when we could have just been using the global store. 

This also decouples Folio-specific select mode functionality from our generic Screen/Header components. This was a bit tricky to figure out but ended up writing a basic portal component that allows us to render things outside of the normal parent / child rendering flow. 

Follow-up PR from here is to remove the remaining uses of `jotai, so that we have exactly one state library being used in energy and can keep things consistent with Eigen. 

https://user-images.githubusercontent.com/236943/221062282-ad850714-be3b-41f5-b653-64bb2b0f69f5.mp4

cc @artsy/mobile-platform 

### PR Checklist

### To the reviewers 👀


src/app/Providers.tsx
```diff
@@ -23,10 +23,8 @@ export const Providers: React.FC<ProviderProps> = ({ children, relayEnvironment
             <RelayProvider relayEnvironment={relayEnvironment}>
               <SafeAreaProvider>
                 <ProvideScreenDimensions>
-                  {/*  */}
```
pvinis, line 26: i had these there cause sometimes prettier is being "smart" and doing `<Provider>{children}</Provider>`, so i used these are guardrails lol. I'm not very attached to them, i was just keeping the diffs clean

src/app/system/store/Models/SelectModeModel.ts
```diff
@@ -20,10 +22,17 @@ export interface SelectModeModel {
   // supporting actions
   setSelectMode: Action<this, this["sessionState"]["isActive"]>
   clearSelectedItems: Action<this>
+
+  // TODO
```
pvinis, line 26: is this still a todo?

src/app/screens/Shows/ShowTabs/ShowDocuments/ShowDocuments.tsx
```diff
@@ -24,49 +25,60 @@ export const ShowDocuments = ({ slug }: { slug: string }) => {
...
-          <ListEmptyComponent text="No documents" mx={SCREEN_HORIZONTAL_PADDING} />
-        }
-      />
-    </TabsScrollView>
+    <>
+      <Portal active={activeTab === "ShowDocuments"}>
+        <SelectMode
+          allSelected={allSelected}
+          selectAll={() => {
+            GlobalStore.actions.selectMode.setSelectedItems({
```
pvinis, line 42: oooo i see. this moved down here. right now this looks stranger to me than the hook, but I'm sure its one of those things that could be easier to read the more you use it.

so the comp is kinda a fragment with the portal bit for select mode, and then the actual list after it. cool cool!

src/app/components/Portal.tsx
```diff
@@ -0,0 +1,44 @@
+import React, { useEffect, createContext, useState } from "react"
+
+interface PortalContextProps {
+  portalChildren: React.ReactNode | null
+  setPortalChildren: (children: this["portalChildren"]) => void
+}
+
+export const PortalContext = createContext<PortalContextProps>({
```
pvinis, line 8: not sure if this is better or not, but what if that portal was the `SelectModePortal`? 🤔  like what if instead of `<Portal><SelectMode {...props} /></Portal>` on the comps, you just had `<SelectModePortal {...props} />`.

pvinis (commented):
nice!

src/app/components/Portal.tsx
damassi: Could be useful for other things, anything really. Just drop the context where you eventually need something to appear and then add a portal somewhere the tree, and suddenly the element is teleported!

src/app/screens/Shows/ShowTabs/ShowDocuments/ShowDocuments.tsx
damassi: Keeping this declarative here was the only way to get rid of the intermediate step, which allowed us to kill that whole select mode state setup file (and other related bits)

PR #546, opened 2023-03-07
refactor(app): Simplify app further 

damassi:
### Description

This trims a ton more fat off the app by again unifying a number of things, instead of duplicating (and triplicating!) everywhere. Summary: 

#### Images 

- We were rendering huge images throughout the app leading to a lot of wasted time, while also threading image size variables throughout every query which created unnecessary and brittle interdependencies.  This removes all of that in favor of a single image definition for general images in our shared image props fragments that's set to be slightly wider than we need, which looks good (we can tweak this easily).
- Additionally, for the large image that we actually cared about rendering large (the zoomable, full-screen one), we fetch that independently, when the user requests it. We didn't need to modify everything for that one spot. 
- Added placeholder grey images around all of the image fetches (so tricky 😓)

#### Category types 

- For things like artist artworks, show artworks, artist documents, show documents, install shots and more, we were creating three separate copies of what was essentially identical code, so i collapsed all of that into things like `ArtworkItem`, `ArtworkList`, `ShowItem` and `ShowList`, etc, and then shared those components throughout. 
- This was made possible by the previous simplification PR that collapsed all of the select mode code into generalized types; once that was done each of these UI items could effectively become "dumb"
- More sharing and simplification can happen here, need to audit a bit more

### Queries
- Remove unnecessary fetches that happen around our Albums. Since that data is stored in local storage, we don't need to make additional requests for data. 
- A little more cleanup/unification can happen here in a follow-up PR, like sharing queries and lists from the artist and artwork list. 

This PR, like the other one, is pretty detailed and so we'll need to QA again and likely do some fixes / follow-up here, but this is a much much better baseline to work from going forward. A lot more cleanup can happen from here, including much easier additional test coverage. 

Need to clean up commits and fix some tests in the AM, and also check offline, but this should be good to pull down and run if someone has a sec. 

cc @artsy/mobile-platform 

#### PR Checklist

### To the reviewers 👀


src/app/components/ImageModal.tsx
```diff
@@ -1,16 +1,53 @@
...
 }
 
-export const ImageModal = ({ isModalVisible, setIsModalVisible, uri }: ImageModalProps) => {
+export const ImageModalQueryRenderer: React.FC<ImageModalProps> = ({ slug, ...props }) => {
+  const { isLoading, data } = useSystemFetchQuery<ImageModalQuery>({
+    query: graphql`
+      query ImageModalQuery($slug: String!, $imageSize: Int!) {
+        artwork(id: $slug) {
+          image {
+            resized(width: $imageSize, version: "normalized") {
```
damassi, line 24: This is the one and only query we need to worry about rendering a really large image

src/app/components/Items/ColumnItem.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+interface ColumnItemProps {
+  index: number
+  numColumns: number
+  children: React.ReactNode
+}
+
+export const ColumnItem: React.FC<ColumnItemProps> = ({ index, numColumns, children }) => {
+  const pl = index % numColumns === 0 ? undefined : 1
+  const pr = index % numColumns === numColumns - 1 ? undefined : 1
```
damassi, line 11: Fixes a lot of messy inconsistent column grid spacing from section to section and unifies it

src/app/components/Lists/ArtworksList.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { GlobalStore } from "app/system/store/GlobalStore"
+import { SelectedItemArtwork } from "app/system/store/Models/SelectModeModel"
+import { usePresentationFilteredArtworks } from "app/utils/hooks/usePresentationFilteredArtworks"
+import { isTablet } from "react-native-device-info"
+
+interface ArtworksListProps {
+  artworks: SelectedItemArtwork[]
+}
+
+export const ArtworksList: React.FC<ArtworksListProps> = ({ artworks }) => {
```
damassi, line 18: Artwork lists can now be shared wherever we need to use / add them

src/app/components/Lists/DocumentList.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { isSelected } from "app/components/SelectMode"
+import { GlobalStore } from "app/system/store/GlobalStore"
+import { SelectedItemDocument } from "app/system/store/Models/SelectModeModel"
+import { getContentContainerStyle } from "app/utils/getContentContainerStyle"
+
+interface DocumentListProps {
+  documents: SelectedItemDocument[]
+}
+
+export const DocumentList: React.FC<DocumentListProps> = ({ documents }) => {
```
damassi, line 14: Same with documents, etc

src/app/components/PageableScreen/PageableLazyScreen.tsx
```diff
@@ -1,18 +1,13 @@
-import { ScrollableScreenEntity } from "app/components/ScrollableScreensView/ScrollableScreensContext"
+import { ScrollableScreenEntity } from "app/components/PageableScreen/PageableScreensContext"
 import { useEffect, useState } from "react"
 import styled from "styled-components/native"
 
-interface ScrollableLazyScreenProps {
+interface PageableLazyScreenProps {
```
damassi, line 5: Renamed to match eigen

src/app/screens/Albums/AlbumTabs/AlbumListImage.tsx
```diff
@@ -1,24 +1,26 @@
...
+  image: SelectedItemArtwork["image"]
 }
 
-export const AlbumListImage = ({ slug, style }: AlbumListImageProps) => {
+export const AlbumListImage = ({ slug, style, image }: AlbumListImageProps) => {
   const placeholderHeight = useScreenDimensions().height / 5
-  const albumListImage = useSystemFetchQuery<AlbumListImageQuery>({
+
+  // Fetch the artwork to validate that it exists, if not, delete from library
+  useSystemFetchQuery<AlbumListImageQuery>({
```
damassi, line 20: I still need to verify this separately

src/app/screens/Albums/AlbumTabs/AlbumListItem.tsx
```diff
@@ -1,13 +1,18 @@
...
-export const AlbumListItem = ({ album }: { album: Album }) => {
+interface AlbumListItemProps {
+  album: Album
+}
+
+export const AlbumListItem: React.FC<AlbumListItemProps> = ({ album }) => {
   const overlapSize: SpacingUnitDSValue = 2
-  const albumItems = album.items ?? []
-  const first3Artworks = albumItems.slice(0, 3)
+  const artworks = usePresentationFilteredArtworks((album.items ?? []) as SelectedItemArtwork[])
```
damassi, line 14: fixes a few different suble bugs. We were evaluating some content based on the items returned from MP, pre filter, so counts were off here and there.

src/app/screens/Artists/ArtistTabs/ArtistArtworks.tsx
```diff
@@ -0,0 +1,49 @@
...
+  const artworks = extractNodes(artworksData.partner?.artworksConnection)
+  const presentedArtworks = usePresentationFilteredArtworks(artworks)
+
+  return (
+    <>
+      <SelectModePortal tabName="ArtistArtworks" items={presentedArtworks} />
+
+      <TabsScrollView>
+        <ArtworksList artworks={presentedArtworks} />
+      </TabsScrollView>
```
damassi, line 31: If we ever need any new screens of this sort, this is the pattern. Add a select mode portal and then the category type, then pass it some items.

src/app/screens/Artwork/ArtworkContent/ArtworkContent.tsx
```diff
@@ -179,35 +179,45 @@ export const ArtworkContent: React.FC<ArtworkContentProps> = ({ artwork }) => {
...
 
+  const resizedImage = {
+    width: screenDimensions.width - space(2),
+    height: screenDimensions.width / (image?.aspectRatio as number) - space(2),
+  }
+
   return (
     <Flex height="100%" ref={touchActivated}>
-      <ImageModal
+      <ImageModalQueryRenderer
```
damassi, line 189: This fetches the large image in the background when the user lands on artwork page. Achieved this by stepping outside of relay's suspense mechanism and instead used `useSystemFetchQuery` under the hood. Its useful to know about!

src/app/screens/Shows/ShowTabs/ShowDocuments.tsx
```diff
@@ -0,0 +1,50 @@
...
+            internalID
+            title
+            filesize
+            publicURL
+          }
+        }
+      }
+    }
+  }
+`
```
damassi, line 50: @admbtlr, @brainbicycle - I know hindsight is 20/20, but given the great similarity between screens in folio, a straightforward component pattern similar to the one above is where we should have started from in the beginning, and when that didn't happen what we ended up with was spiraling, unnecessary complexity that extended out into pretty much every corner of the app, and grew and grew.

The thing about Folio is: the designs show functionality that's fairly straightforward and simple. In most cases, something that looks simple, is. 

We can retro this once thing are shipped, but as a team we need to figure out how to elevate PR review a bit on projects, and especially greenfield ones. For new apps in particular, getting the wider teams eyes on some of the patterns might be good to check assumptions, especially as projects drag on.

admbtlr:
We tried using a single, large image size everywhere to make it easy in offline mode. How does this PR impact that? Do we now need to download multiple sizes of each image?

damassi:
For an artwork, there's two images in the app -- one that's super large (zoomable) and then one that's medium sized, for thumbnails, that loads real quick. 

When offline, the medium-sized one loads real fast (and more performant, in the UI, because its smaller), and then if the user wants to see a full screen zoom, that one then loads, if need be. 

Spreading around the single large image container was slightly overthinking it given the UI impact and the overhead of having to thread code, pass arguments, and complexify queries. It also broke the principle of 'use whats appropriate'; a tiny thumbnail on a tiny phone doen't need a large image for an offline feature that hasn't been written yet. If we try to always do whats appropriate, things will work themselves out!

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 19f6b5faffd5da333e0819d1016d29ef220d4131
</p>

PR #556, opened 2023-03-10
feat(albums): Add album item existence validation hook

damassi:
### Description

This adds a new `useValidateAlbumItems` hook which will execute when the album loads and validate whether items in the album still exist; if they don't, the items are removed and an alert is shown. 

Album items are rendered from local-storage and the UI responds immediately, and we push the fetching to the background. 

Will follow-up in the AM with some tests but wanted to open up for review. 

cc @artsy/mobile-platform 

### PR Checklist


src/screens/Albums/useValidateAlbumItems.tsx
```diff
@@ -0,0 +1,53 @@
...
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [data])
+}
+
+const showAlertMessage = debounce(() => {
+  Alert.alert(
+    "Items from this collection have been deleted in CMS.",
+    "Your album has been updated."
+  )
+}, 1000)
```
damassi, line 53: We fire all three tab checks upon entering the view; this ensures that only one alert message is shown.

brainbicycle (approved):
looking good!

PR #563, opened 2023-03-14
refactor(app): Refactor and Simplify Screen, Replace jotai state with a context, add Toasts

damassi:
### Description

Pretty much completely stripped down the `Screen` abstraction to the bare essentials, removed all confusing state patterns, deleted unused components, and eliminated anything that isn't being currently used in Folio. There could be things I'm missing here, but I went through every screen and surprised myself again and again at what a simple flex component can do, when compared with everything else that was there before. 

**Flow**: 
- Created new Screen2 folder
- Created a new Screen2.Whatever component, and referred to Screen.Whatever
- Picked out _only_ UI aspects from each component, the bare essentials, no state or any behavior at all
- Swapped out Screen.Whatever with Screen2.Whatever 
- Does it look good? Yep
- Repeat for every component in the Screen directory
- For anything that relied on shared state to render certain UI elements, I removed and rewrote so that it would behave statically 
- Aside from triggering the animation in the header, no state is necessary (at all) and i'm very unsure what was going on there.  
- After checking everything, deleted `components/Screen`, and then renamed `components/Screen2` to `Screen` so that we could keep the git log 
- Following this flow, all complexity could be removed and unless i'm missing something obvious, it behaves in the exact same way as before 🤔 🤔 

Also, search pills stuff was rewritten to remove Jotai state, and adds the ability to search through albums by cross-referencing returned artwork internalIDs, and checking if said ids are in any albums; if so, return the result. 

And cleaned up some other random UX inconsistencies along the way. 

Toast support was added: 

https://user-images.githubusercontent.com/236943/225456869-2737321e-2abd-4cb3-abc7-128fbed030a7.mp4

Lastly, all patch-packages were removed. 

There's one last jotai state piece being used, in the `fileManager`; will remove in a follow-up PR. 

cc @artsy/mobile-platform 

### PR Checklist

- [X] I have documented any **follow-up work** that this PR will require, or it does not require any.


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against f08fb17dd4b029bed3da137315a8753be513dfca
</p>

damassi:
@brainbicycle pulled this down and gave it a spin, looked good except header animation. That's been fixed so gonna merge and send out beta and can deal with any fallout from there 👍

PR #565, opened 2023-03-16
feat(animation): Add moti, swap out animations

damassi:
### Description

This swaps out our imperative animation code with [Moti](https://moti.fyi/), an amazingly declarative animation API built on top of reanimated. Now we can just describe what we want to do in plain language, rather than building everything from scratch. Much simpler! 

Worth checking out the deleted files as a comparison. 

Thanks for the heads up @MounirDhahri 👍 

Follow-up here is an RFC to bring this over to Eigen. 

cc @artsy/mobile-platform 

### PR Checklist


MounirDhahri (approved):

MounirDhahri:
This is really great!

MounirDhahri:
I think it would be great to have some kind of hidden route to include those animations so everyone is aware of their existence - a storybook like implementation 😄

PR #568, opened 2023-03-17
fix(offline): Fix sync validation

damassi:
### Description

Continues offline investigations. This fixes the sync error validation by not retrying the whole category, but rather only refetching the exact query that failed. This will greatly improve perf, especially for us on staging testing large partners which often have 404s. 

Also comments out the retry logic around file downloads as I think that might be causing issues and possibly overloading the system, since each error could be invoked 3 times, * n number of errors, within the process of actually fetching files. 

Lastly, clears the file cache when we start the sync just so we're sure we're now filling up the device. 

Lets see how this works tomorrow. 

cc @artsy/mobile-platform


src/system/sync/fileCache/downloadFile.ts
```diff
@@ -73,8 +74,10 @@ export const downloadFileToCache = async ({ url, type, accessToken }: DownloadFi
           `[fileCache] Error downloading file: ${error}. Retrying [${errorRetryAttempt}]: ${url}`
         )
 
+        // TODO: Investigate whether this might be causing problems. Comment
+        // out for now.
         // Retry download
-        await startDownload()
+        // await startDownload()
```
damassi, line 80: @brainbicycle - Comment out the file download retry loop for now. This is separate from refetching queries that fail. This is like: process a) query returns image url; process b) download image from url. Did it fail? Retry, internally to the function and separate from the overall syncManager.  

This worked really good for smaller partners, but when running against partners with thousands of images, this started to get congested here and there. If something taps out and the file fails, it doesn't break anything, so not the end of the world. Will need to rethink how to do this in a more organized fashion. I'll file in JIRA for now.

PR #572, opened 2023-03-18
chore(test): Add missing test and comment out download retry as test

damassi:
### Description

This comments out the download retry for the time being, to see how things work without it. 

Also adds another test. 

cc @artsy/mobile-platform


src/system/sync/syncManager.ts
```diff
@@ -178,8 +178,9 @@ export function initSyncManager({
       syncInstallShots,
       syncDocuments,
 
+      // TODO: Reenable
       // Retry download errors caught above 3 times
-      retryFileDownloadsForErrors,
+      // retryFileDownloadsForErrors,
```
damassi, line 183: Stalled out at 58% Wanna see how things work without this on, just to isolate possibilities.

damassi: Thought that said, since this is now using the same setup as everything else, it shouldn't quite matter -- its just another queue item 🤔. Frustrating. 

In any case, we should raise the possibility of Gemini throttling to #product-velocity, see if they might have any insights.

PR #573, opened 2023-03-18
temp(debugging): Add offline developer debug tooling

damassi:
### Description

This adds some developer tooling while we debug the image download issue in offline. In the offline mode, if a developer, you should see a new JSON tree below the clear cache button. As data is downloaded, the tree is uploaded. 

At first glance and i'm seeing some keys with no values! So I think that's a significant lead. 

https://user-images.githubusercontent.com/236943/226084050-c8f11dc3-c200-44ec-803c-51d5a979d9f6.mp4


src/system/sync/syncManager.ts
```diff
@@ -161,22 +167,22 @@ export function initSyncManager({
...
-      syncShowArtworksQuery,
-      syncShowInstallsQuery,
-      syncShowDocumentsQuery,
+      // syncArtistShowsQuery,
+      // syncArtistDocumentsQuery,
+      // syncShowTabsQuery,
+      // syncPartnerShowTabsQuery,
+      // syncShowArtworksQuery,
+      // syncShowInstallsQuery,
+      // syncShowDocumentsQuery,
```
damassi, line 176: Don't need these items while you debug

src/system/sync/syncManager.ts
```diff
@@ -237,6 +248,7 @@ export function initSyncManager({
 
     const { results } = await PromisePool.for(artistSlugs)
       .onTaskStarted(reportProgress("Syncing artist artworks"))
+      .withConcurrency(50)
```
damassi, line 251: Bumped the concurrency way up on MP requests as i think it can handle it while we test. This is separate from the concurrency while we're hitting gemini, which seems to be more sensitive.

PR #578, opened 2023-03-23
fix: sync crash and some debug tools

brainbicycle:
This PR resolves [] 

### Description

Removes null values from results before assigning them to the sync object, before if a query failed with a nonstandard graphql error (e.g. html page showing a server error) it would be null in the object and crash when attempting to save to the cache. 

Adds a viewer for the url map, adds some files to test image sync without running the other steps, should be removed.

Follow-ups:
- remove the test files and find a better fast feedback loop that doesn't require inclusion in the project
- clean up the unnecessary logging
- remove the unnecessary url map clearing, if it is actually a problem see if there is another way

### PR Checklist

### To the reviewers 👀


src/screens/Settings/OfflineModeSettings.tsx
```diff
@@ -74,6 +77,7 @@ export const OfflineModeSettings = () => {
 
   const handleClearFileCache = async () => {
     await clearFileCache()
+    await clearUrlMap()
```
brainbicycle, line 80: this is not necessary, should be removed. I am seeing the urlMap not being cleared out / still loading after tapping the clear cache button which in theory is supposed to delete the directory. Just occurred to me while writing, maybe the directory is cleared but it needs to be cleared from memory.

src/system/sync/fileCache/urlMap.ts
```diff
@@ -27,6 +27,18 @@ export const saveUrlMap = async () => {
...
+  try {
+    const path = PATH_CACHE + "/urlMap.json"
+    if (!(await exists(path))) {
+      return
+    }
+    await unlink(path)
+  } catch (error) {
+    console.log("[fileCache] Error deleting urlMap:", error)
+  }
+}
```
brainbicycle, line 40: remove this

src/system/sync/fileCache/downloadFileToCache.ts
```diff
@@ -60,6 +64,8 @@ export const initDownloadFileToCache = ({ onFileDownloadError }: InitDownloadFil
         try {
           await unlink(getFilePath({ filename: urlMap[url], type }))
         } catch (error) {
+          console.log("imagesync unlink error", error)
```
brainbicycle, line 67: remove these logs or make togglable

src/system/sync/syncManager.ts
```diff
@@ -581,19 +583,26 @@ const parsers = {
...
+    const rawImageUrls = [
+      ...(syncResults.artworkQuery ?? []).flatMap((artworkContent) => {
+        if (artworkContent?.artwork?.image?.resized?.url) {
+          return artworkContent.artwork.image.resized?.url
+        } else if (artworkContent?.artwork?.artist?.imageUrl) {
+          return artworkContent.artwork.artist.imageUrl
+        } else {
+          return null
+        }
+      }),
```
brainbicycle, line 595: make sure you didn't break anything here ☝️

src/screens/Settings/OfflineModeSettings.tsx
```diff
@@ -148,7 +151,20 @@ export const OfflineModeSettings = () => {
             Clear cache
           </Button>
 
+          <Button
+            block
+            onPress={async () => {
+              await loadUrlMap()
+              const rawURLMap = getURLMap()
+              setURLMap(rawURLMap)
```
brainbicycle, line 159:
```suggestion
              const updatedURLMap = getURLMap()
              setURLMap(updatedURLMap)
```

PR #591, opened 2023-04-12
fix(webview): Fix webview crashes on Android

damassi:
This PR resolves [MOPLAT-683]

### Description

Struggled with this for a while but in the end just settled on simplifying things. 

For some reason, including any other component in the context of the webview component -- no matter what it was -- would lead to a subsequent crash on android. So just left the webview alone; the screen loads fast enough that we don't need to worry about tracking progress and so on. 

cc @artsy/mobile-platform 

### PR Checklist

[MOPLAT-683]: https://artsyproduct.atlassian.net/browse/MOPLAT-683?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/screens/Artwork/ArtworkWebView.tsx
```diff
@@ -1,90 +1,22 @@
...
-          source={{ uri }}
-          sharedCookiesEnabled
-          onLoadProgress={(e) => setLoadProgress(e.nativeEvent.progress)}
-          onLoad={() => setIsProgressBarVisible(false)}
-        />
-        {isProgressBarVisible && (
-          <Flex height="100%">
-            <ProgressBar progress={loadProgress} />
-          </Flex>
-        )}
```
brainbicycle, line 32: I would prefer not to remove this if it is avoidable, do we an idea why having an additional component causes a crash?

brainbicycle (approved):

src/screens/Artwork/ArtworkWebView.tsx
brainbicycle: not strong enough to block, we can merge and take a look later

damassi: Ok i fixed it! in the end it was a css issue.

PR #613, opened 2023-05-04
feat(palette-mobile): Point <Screen /> at palette-mobile

damassi:
Related:
- https://github.com/artsy/palette-mobile/pull/98

This PR resolves [MOPLAT-759]

### Description

This updates energy to use palette-mobile's new updated screen component. 

One part that wasn't touched in this was the one component that uses [Screen.AnimatedTabHeader](https://github.com/artsy/energy/blob/main/src/components/Tabs/TabsWithHeader.tsx#L38) which relies on the tabs to return scroll position. Will follow up here to make things a little more generic so that we can share Screen.Header and still animate.  

cc @artsy/mobile-platform 

### PR Checklist

[MOPLAT-759]: https://artsyproduct.atlassian.net/browse/MOPLAT-759?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/utils/hooks/useIsDarkMode.ts
```diff
@@ -0,0 +1,6 @@
+import { GlobalStore } from "system/store/GlobalStore"
+
+export const useIsDarkMode = () => {
+  const isDarkMode = GlobalStore.useAppState((state) => state.devicePrefs.colorScheme === "dark")
+  return isDarkMode
+}
```
gkartalis, line 6: 👏

src/Navigation.tsx
```diff
@@ -65,7 +67,7 @@ export const Main = () => {
         <StackNav.Navigator screenOptions={{ headerShown: false }} initialRouteName="HomeTabs">
           {AuthNavigation({ isLoggedIn, selectedPartner })}
 
-          {isLoggedIn && selectedPartner && (
+          {!!isLoggedIn && !!selectedPartner && (
```
gkartalis, line 70: I love these autofixes slowly getting their way into our code 🧹⭐️

src/screens/Albums/CreateOrEditAlbum/CreateOrEditAlbumChooseArtworks.tsx
```diff
@@ -79,7 +81,7 @@ export const CreateOrEditAlbumChooseArtworks = () => {
           }}
         />
 
-        <Screen.BottomView>
+        <Screen.BottomView darkMode={isDarkMode}>
```
gkartalis, line 84: **question:** Why is darkMode needed to be passed on the bottomView only? Wouldn't be more straightforward to pass it to the `<screen>` component? 🤔

gkartalis: Or we don't need it there and I am missing something

gkartalis (approved):
will run locally to also check if there is something obvious that we missed but looks great!

PR #588, opened 2023-04-11
feat: offline improvements

gkartalis:
This PR resolves [MOPLAT-677] 

### Description

- [x] adds keep-awake library to keep screen active when downloading 
- [x] add spinner when syncing

### PR Checklist

- [x] I have tested my changes on **iOS** and **Android**.
- [x] I have included **screenshots** or **videos**, or I have not changed the UI.
- [x] I have added **tests**, or my changes don't require any.
- [x] I added an **[app state migration]**, or my changes do not require one.
- [x] I have documented any **follow-up work** that this PR will require, or it does not require any.

### To the reviewers 👀

- [ ] I would like **at least one** of the reviewers to **run** this PR on the simulator or device.

Need help with something? Have a look at our [docs], or get in touch with us.

[app state migration]: ../blob/main/docs/adding_state_migrations.md
[feature flag]: ../blob/main/docs/developing_a_feature.md
[docs]: ../blob/main/docs/README.md

[MOPLAT-677]: https://artsyproduct.atlassian.net/browse/MOPLAT-677?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #626, opened 2023-05-15
refactor(offline): If image already downloaded, skip downloading it again

damassi:
This PR resolves [MOPLAT-773]

### Description

This adds some functionality to check to see if we've already downloaded the image; if so, skip downloading it again. 

Also removes our dynamic UUID-generated files which created a unique file name on disk for every file, every time we downloaded, in favor of consistent IDs, keyed from cloudfront / gemini. 

And fixes an issue that was leading to app stability from [here](https://github.com/artsy/energy/pull/588), which accidentally remounted the network listener every time we visited a new subview. Whenever it would mount, it would trigger a load on the offline cache and blow away the view tree. 

### PR Checklist

[MOPLAT-773]: https://artsyproduct.atlassian.net/browse/MOPLAT-773?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/Tabs/TabScreen.tsx
```diff
@@ -1,13 +1,13 @@
...
 import { SuspenseWrapper } from "system/wrappers/SuspenseWrapper"
 
+// TODO: Wire this up to checks from the relay store
+// import { OfflineLoadingWrapper } from "system/sync/components/OfflineLoadingWrapper"
+
 export const TabScreen: React.FC = ({ children }) => {
   return (
     <RetryErrorBoundary withoutBackButton>
-      <SuspenseWrapper withTabs>
-        <OfflineLoadingWrapper>{children}</OfflineLoadingWrapper>
```
damassi, line 9: We should return to this once we can connect it better to the relay store record queue

brainbicycle (approved):

src/system/sync/syncManager.ts
```diff
@@ -167,8 +166,9 @@ export function initSyncManager({
   const startSync = async () => {
     updateStatus("Starting sync")
 
+    // TODO: Do we actually need this?
     // Be sure we're starting from scratch
-    await clearFileCache()
+    // await clearFileCache()
```
damassi, line 171: We were really making testing this offline image stuff very hard on ourselves 😭

src/components/Tabs/TabScreen.tsx
brainbicycle: Does this mean the loading state when open the app offline is gone?

damassi: Yup, for now. The way we're doing it has to be rethought a bit (maybe by using the global store, outside of this file?) 

Will definitely add this back.

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 3505b9dd267ef24d835d77f3cffbbd88a7a73bea
</p>

src/components/Tabs/TabScreen.tsx
damassi: Added in https://github.com/artsy/energy/pull/626/commits/1b8fe9298a037e3af573d37a14520b2f8e4e8ea7

PR #647, opened 2023-06-05
build(android): remove hermes residue

gkartalis:
This PR resolves [] 

### Description

Cleaning up hermes from folio since it is not setup correctly and might cause slower performance on Android

### PR Checklist

### To the reviewers 👀


ios/Podfile.lock
```diff
@@ -695,7 +695,7 @@ SPEC CHECKSUMS:
   boost: a7c83b31436843459a1961bfd74b96033dc77234
   BVLinearGradient: 34a999fda29036898a09c6a6b728b0b4189e1a44
   CocoaAsyncSocket: 065fd1e645c7abab64f7a6a2007a48038fdc6a99
-  DoubleConversion: 831926d9b8bf8166fd87886c4abab286c2422662
+  DoubleConversion: 5189b271737e1565bdce30deb4a08d647e3f5f54
```
gkartalis, line 698:
```suggestion
  DoubleConversion: 831926d9b8bf8166fd87886c4abab286c2422662
```

android/app/build.gradle
```diff
@@ -88,7 +88,7 @@ import com.android.build.OutputFile
  */
 
 project.ext.react = [
-    enableHermes: true,  // clean and rebuild if changing
```
damassi, line 91: Ah, so this was never set up correctly? Interesting!

gkartalis: Yep, so this by itself can become problematic. 

It was a gotcha when we did that on eigen. This by itself will not convert the code to `bytecode` and we will need to do the extra steps that we did in eigen.

If we leave this here makes the app actually more slow 😅 realized it when we first build `eigen` with hermes enabled but without the extra build steps. 

We need to do something like:
[Bundle JS](https://github.com/artsy/eigen/blob/main/scripts/hermes-android-bundle.sh) 
[Bundle bytecode](https://github.com/artsy/eigen/blob/main/scripts/hermes-android-bundle-bytecode.sh)
[Create the sourcemaps](https://github.com/artsy/eigen/blob/main/scripts/hermes-android-sourcemap.sh)

damassi: Ah. For some reason I thought that Energy was all setup for Hermes from the get-go, via the project template. Was just the flag 😄 But cool that we now know the steps

PR #761, opened 2023-11-08
build(deps): Replaces expo and expo-mail-composer and uses react-native-mail instead

gkartalis:
This PR resolves [PHIRE-426] [PHIRE-467] 

### Description

Removes `expo` and `expo-mail-composer` and replaces it with `react-native-mail`.

Nothing changes in the implementation.

We have the same behavior as before meaning that iOS mail works as before and android as well -

Android cannot attach images and we see the `[OBJ]` square on the mail body **as before**.

But the huge win is that we got rid of expo and its underlying dependencies 

### PR Checklist

### To the reviewers 👀

[PHIRE-426]: https://artsyproduct.atlassian.net/browse/PHIRE-426?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[PHIRE-467]: https://artsyproduct.atlassian.net/browse/PHIRE-467?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


damassi (approved):
Worth noting that the current email feature used here is insufficient anyways. One of the big issues is that we don't have the ability to launch email in default browser, that kind of thing. So we should keep that in mind!

gkartalis:
> Worth noting that the current email feature used here is insufficient anyways. One of the big issues is that we don't have the ability to launch email in default browser, that kind of thing. So we should keep that in mind!

The previous implementation wasn't working on android - wasn't able to add the images to the body so it was incomplete eitherway. 

Worth looking into `react-native-mail` as a replacement that might be better.

src/screens/Artwork/useMailComposer.ts
```diff
@@ -1,221 +1,231 @@
...
+    console.log(
+      "[useMailComposer] Sending email:",
+      {
+        subject,
+        ccRecipients,
+      },
+      body
+    )
+  }
+}
```
damassi: This whole file.... 😅

ios/Energy/AppDelegate.h
```diff
@@ -1,8 +1,7 @@
 #import <React/RCTBridgeDelegate.h>
-#import <Expo/Expo.h>
 #import <UIKit/UIKit.h>
 
-@interface AppDelegate : EXAppDelegateWrapper <UIApplicationDelegate, RCTBridgeDelegate>
```
brainbicycle, line 5: should this be UIResponder 🤔 will check on a fresh project quickly

brainbicycle: yep

brainbicycle (approved):
🙌 awesome

PR #800, opened 2023-12-27
chore: analytics init

brainbicycle:
This PR resolves [PHIRE-535] 

### Description

Adds a basic setup for tracking segment analytics using eigen as a guide: 
- adds react-tracking, cohesion, @segment/analytics-react-native, and associated deps and follows the docs for setup for those dependencies
- adds a analytics provider and infrastructure for tracking based on Eigen's tracking infra, only major difference is any legacy code related to non-cohesion events was removed 

@damassi This works I believe but open to something simpler if we want, very much does not have to be merged as is. I will leave it here if you want to work off it. 

### PR Checklist

### To the reviewers 👀

[PHIRE-535]: https://artsyproduct.atlassian.net/browse/PHIRE-535?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/utils/track/SegmentTrackingProvider.ts
```diff
@@ -0,0 +1,52 @@
+import { createClient, SegmentClient } from "@segment/analytics-react-native"
```
damassi, line 1: @brainbicycle - i think all the stuff from this file on down can be omitted in favor of some simpler patterns from Force et al. A lot of this code is old school eigen stuff; we can keep all of our new code 1:1 across environments (except for the RN init part)

damassi: Something like 
- https://github.com/artsy/volt-v2/blob/main/src/system/providers/AnalyticsProvider.tsx and 
- https://github.com/artsy/volt-v2/blob/main/src/system/hooks/usePageViewTracking.ts (though tapped into Energy's router) 

And then from there, we just use `useTracking` from react-tracking. 

Happy to take things from here if you don't feel like working through this -- Also, can understand if we might want to keep things 1:1 with eigen.

brainbicycle: yeah I am a bit torn, I think in an ideal world if we were going to adopt a simple pattern we would adopt it in both places and I am a bit worried about drift between the codebases. But maybe the analytics code will be simple enough that it is not a major concern.

brainbicycle: If you'd like you are welcome to take it from here and adopt the force style. I don't want to force legacy patterns on folio either we can use it as a guide to refactor eigen in future. Where this is at now: segment is integrated and initializing just point to the old folio source in segment we will want to set up the new sources and update the initialization code accordingly. The visualizer is also pretty nice for analytics testing so might be nice to bring that over but a nice to have for sure.

src/utils/track/SegmentTrackingProvider.ts
```diff
@@ -0,0 +1,52 @@
...
+      } else {
+        // TODO: Maybe we should have this for testing purposes
+        // visualize("Track", action, info, visualizeDevToggle)
+        analytics.track(action, info as any)
+      }
+      return
+    }
+
+    console.warn("oh wow, we are not tracking this event!! we should!", { info })
+    assertNever(info)
```
brainbicycle, line 50: I am not sure this works 🤔 maybe was removed in later rn versions and we didn't notice? Will check eigen

src/utils/track/SegmentTrackingProvider.ts
brainbicycle: also just a thought we may need some mechanism for tracking page views that don't go through the router, don't know if it is true of any pages today in energy but generally on mobile not guaranteed that a page goes through the router to be presented.

package.json
```diff
@@ -177,6 +183,9 @@
     "sha1-sum": "^1.0.0",
     "typescript": "^4.9.4"
   },
+  "resolutions": {
+    "core-js": "3.21.1"
```
brainbicycle, line 187: I am not sure if this is something that needs to be addressed in cohesion but installing I got version "3.26.1" of core-js installed as a dependency and a red screen on start up, might need to either update the required version in cohesion to be more specific or update cohesion to be compatible with later versions 🤔

ios/Energy.xcodeproj/project.pbxproj
```diff
@@ -661,7 +661,7 @@
 			buildSettings = {
 				ALWAYS_SEARCH_USER_PATHS = NO;
 				CLANG_ANALYZER_LOCALIZABILITY_NONLOCALIZED = YES;
-				CLANG_CXX_LANGUAGE_STANDARD = "gnu++14";
```
brainbicycle, line 664: change came with upgrade to Xcode 15.1

.vscode/settings.json
```diff
@@ -1,6 +1,6 @@
 {
   "editor.codeActionsOnSave": {
-    "source.fixAll.eslint": true
```
brainbicycle, line 3: unrelated to analytics this property was renamed in newer versions of vscode

.circleci/config.yml
```diff
@@ -147,6 +147,7 @@ commands:
       - run:
           name: Build App
           command: ./scripts/ci/ci-android
+          no_output_timeout: 20m
```
brainbicycle, line 150: unrelated to analytics but android builds are sometimes timing out

ios/Podfile
```diff
@@ -13,6 +13,8 @@ ENVFILES = {
   'Release' => '$(PODS_ROOT)/../../.env',
 }
 
+ENV['SWIFT_VERSION'] = '5'
```
brainbicycle, line 16: initial analytics pods failed to install because swift version was not specified

src/utils/track/SegmentTrackingProvider.ts
damassi: Ahh that's a good point, though from what I can tell in eigen, even the modal views are handled by react-navigation. That said, not sure if that's considered a route, though i'm pretty sure it is. Might be something to keep an eye on. 

And yeah, also a bit torn about the approach here, but yeah. I _think_ this is all that we need for now, and we can skip the old eigen patterns.

damassi: We can go ahead and merge and I can update if things change 👍

damassi (approved):
Thanks so much for the setup 👍 Can return to this to refactor if necessary.

PR #801, opened 2024-01-02
feat: android email images as pdf 

brainbicycle:
This PR resolves [PHIRE-519] 

### Description

After doing some digging on inline images in html emails it seems it is not reliably possible on Android.
https://stackoverflow.com/questions/7787171/sending-html-email-in-android-using-table-etc-is-there-really-no-relativel

This PR is an alternative, we would use the same html body we use on iOS to create a pdf of the images and artwork metadata and send that as an attachment on Android only. 

- Adds dependency react-native-html-to-pdf which takes html content and creates a pdf, only minimal changes were needed to get it working but have not tested thoroughly.

Open to suggestions, alternatives, complete rejection :). 

Here is an example of how it looks in the generated pdf:

Screenshot 2024-01-02 at 3 37 30 PM

### PR Checklist

### To the reviewers 👀

[PHIRE-519]: https://artsyproduct.atlassian.net/browse/PHIRE-519?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/screens/Artwork/useMailComposer.ts
```diff
@@ -28,7 +29,12 @@ export const useMailComposer = () => {
         .replace("$title", title ?? "")
         .replace("$artist", artistNames ?? "")
 
-      const body = getArtworkEmailTemplate({ artwork: firstSelectedItem, emailSettings })
+      const htmlContent = getArtworkEmailTemplate({ artwork: firstSelectedItem, emailSettings })
+
+      const body = Platform.OS === "ios" ? htmlContent : "Please see attached artworks."
```
brainbicycle, line 34: can be whatever but figured something simple to point people to the attachment.

damassi (approved):
Not so bad in terms of implementation! And a really good compromise solution. Thanks so much 🙏

PR #814, opened 2024-01-12
fix: email statuses

gkartalis:
This PR resolves [] 

### Description

Email wasn't respecting the statuses and as a result it was returning both success and error at the same time.

Reason for that was that we didn't take into account the event status strings. 

This PR fixes that.

Need to thouroughly test it out to understand when it works and when it doesn't ~ I think there are some issues on iOS 17 for example wasn't able to see the actual html template.

### PR Checklist

### To the reviewers 👀


src/screens/Artwork/useMailComposer.ts
```diff
@@ -38,31 +100,14 @@ export const useMailComposer = () => {
...
-
-      if (!mailError) {
-        toast.show({
-          title: "Email sent.",
-          type: "info",
-        })
-      }
-
-      mailError = ""
+      emailComposer({
```
damassi, line 103: Ok that this isn't `await`'d?

damassi (approved):
One q, but otherwise tyyy

src/screens/Artwork/useMailComposer.ts
```diff
@@ -8,6 +8,68 @@ import { GlobalStore } from "system/store/GlobalStore"
...
+
+const emailComposer = async ({
+  subject,
+  ccRecipients,
+  body,
+  isHTML,
+  attachments,
+  toast,
+}: EmailComposerProps) => {
+  await Mailer.mail(
```
gkartalis, line 28:
```suggestion
  Mailer.mail(
```

src/screens/Artwork/useMailComposer.ts
```diff
@@ -8,6 +8,68 @@ import { GlobalStore } from "system/store/GlobalStore"
...
+interface EmailComposerProps {
+  subject: string
+  ccRecipients?: string[]
+  body: string
+  isHTML: boolean
+  attachments?: any[]
+  toast: ReturnType<typeof useToast>["toast"]
+}
+
+const emailComposer = async ({
```
gkartalis, line 20:
```suggestion
const emailComposer = ({
```

src/screens/Artwork/useMailComposer.ts
gkartalis: no need to await that, will also remove the asyn await from the emailComposer func

PR #825, opened 2024-01-23
fix(albums): Always show add works album

damassi:
This PR resolves [AMBER-417]

### Description

This fixes the issue where the add artworks button disappears after adding an item. Now we always keep it on the screen:

https://github.com/artsy/energy/assets/236943/e1834ee3-a326-44a6-95d8-f277c8d92a08

[AMBER-417]: https://artsyproduct.atlassian.net/browse/AMBER-417?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/screens/Albums/CreateOrEditAlbum/CreateOrEditAlbum.tsx
```diff
@@ -124,7 +124,6 @@ export const CreateOrEditAlbum = () => {
     )
   }
 
-  const showAddMessage = isSelectModeActive || !artworksToAdd?.length
```
jonallured, line 127: Hrm, what was the previous thought here - were we trying to hide this while adding a work? Or was this just a miss?

damassi: I think it was an oversight of some kind, but not too sure

damassi:
Gonna merge 👍

PR #830, opened 2024-01-24
[WIP] feat: Wire up tracking AMBER-434

jonallured:
This PR adds some tracking to Folio - opening before it's done for some feedback from @damassi and @brainbicycle (or anybody really 🤗 ) especially around some questions I left as comments.

https://artsyproduct.atlassian.net/browse/AMBER-434

/cc @artsy/amber-devs


damassi:
@jonallured - i think you forgot to push yr latest commit? not seeing any qs / comments

src/screens/Auth/Login.tsx
```diff
@@ -75,6 +78,9 @@ export const LoginScreen = () => {
       }
 
       if (success) {
+        // where can i get userId?
+        const userId = "abc123"
```
brainbicycle, line 82: on the auth model in the global store: https://github.com/artsy/energy/blob/4a7962f8331d055ce57fbe05925d2e3eac8874a6/src/system/store/Models/AuthModel.ts#L17

example access would be something like this:
https://github.com/artsy/energy/blob/4a7962f8331d055ce57fbe05925d2e3eac8874a6/src/utils/hooks/useIsOnline.ts#L4

jonallured: I'm not sure this will work because this tracking call is happening inside of the Formik `onSubmit` handler. When I attempt to access the `GlobalStore` I get a message like this:

```
[eslint] React Hook "GlobalStore.useAppState" is called in function "onSubmit" that is neither a React function comp...
```

Which makes me wonder if maybe this is the wrong place to do this tracking call? What do you think?

PR #837, opened 2024-01-26
feat(analytics): Implement tracking 

damassi:
### Description

This PR simplifies / refactors the existing analytics implementation, wires up screen tracking, adds a 1st-launch pattern, and fills out other events. 

#### Screen Tracking 

To add screen tracking, import the `useTrackScreen` hook:

```tsx
const App = () => {
  useTrackScreen("MyScreen")
  ...
}
```

The arguments that are passed into `useTrackScreen` are automatically inferred from our react-navigation stack-based route keys; and if any non-stack route needs to be added (such as tab views), it can be added to the `Screens` type in that file. Since this approach (and the underlying analytics setup) are just a few lines of code, we might consider bringing this over to Eigen.

#### Event Tracking  

To track an event, use the `useAppTracking` hook:

```tsx
const App = () => {
  const { trackLoggedIn } = useAppTracking()
  
  return <button onClick={trackLoggedIn} />
}
```

This PR also adds a "first launch?" implementation to Energy, to track when the user boots the app for the first time. Its much simpler than Eigen's and just relies on local-storage. Local storage is only blown away if the user deletes the app, so we can safely assume whats read from there as valid. This has been added to the `SystemModel` under the `launchCount` key. (Changes to our models are automatically persisted to localStorage). 

https://github.com/artsy/energy/assets/236943/f509571d-8298-4b95-bd6b-d733ec54987f

Follow-up PR will 
- adjust said screens type to account for Louis's cohesion events
- wire up some misc details in the tracking that need some clarifying (per todos in code)

cc @artsy/amber-devs


index.js
```diff
@@ -2,14 +2,9 @@ import "react-native-gesture-handler"
...
 import { AppRegistry } from "react-native"
-import { enableFreeze } from "react-native-screens"
 import { name as appName } from "./app.json"
 import { App } from "./src/App"
 
 global.__TEST__ = false
 
-// Prevent React component subtrees from rendering.
-// This is still experimental so we can try it out and disable it later if it's unstable
-enableFreeze(false)
```
damassi, line 13: Noticed a bug from this today, in the pull to refresh.

src/screens/Settings/navigation.tsx
```diff
@@ -30,12 +30,15 @@ export const SettingsNavigation = () => {
       <StackNav.Screen name="OfflineModeSettings" component={OfflineModeSettings} />
 
       <StackNav.Screen name="EmailSettings" component={EmailSettings} />
-      <StackNav.Screen name="MultipleArtworksAndArtists" component={MultipleArtworksAndArtists} />
       <StackNav.Screen
-        name="MultipleArtworksBySameArtist"
+        name="EmailSettingsMultipleArtworksAndArtists"
```
damassi, line 34: This is where the screenTracking keys are inferred from and piped through into:
 
Screenshot 2024-01-25 at 5 36 19 PM

src/system/analytics/AnalyticsProvider.tsx
```diff
@@ -0,0 +1,8 @@
+import track from "react-tracking"
+import { trackSegmentEvent } from "system/analytics/trackSegmentEvent"
+
+export const AnalyticsProvider = track(undefined, {
+  dispatch: trackSegmentEvent,
+})(({ children }: any) => {
+  return children
+})
```
damassi, line 8: 🤷 Not sure exactly whats going on in Eigen, but we can certainly refactor

src/system/hooks/useAppTracking.ts
```diff
@@ -0,0 +1,136 @@
...
+      }
+
+      const event = {
+        action: "first_user_install",
+      }
+
+      trackEvent(event)
+    },
+
+    trackLoginSuccess: () => {
```
damassi, line 40: The following events haven't been wired into the app, just stubbed out here. That's for the next PR.

src/system/store/Models/SystemModel.ts
```diff
@@ -17,6 +21,12 @@ export const getSystemModel = (): SystemModel => ({
     navigationHistory: {},
   },
 
+  launchCount: 0,
+
+  incrementLaunchCount: action((state) => {
+    state.launchCount += 1
+  }),
```
damassi, line 28: We might want to consider refactoring the first launch code in eigen too. Doesn't need to be more than this (I don't think)

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 629bd2a30c0b189e6268edcfe6ed8f07f160a687
</p>

src/system/hooks/useTrackScreen.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  const activeTab = (() => {
+    try {
+      // Calling hook conditionally to guard against being outside of TabContext
+      // eslint-disable-next-line react-hooks/rules-of-hooks
+      return useFocusedTab()
+    } catch {
+      return null
+    }
+  })()
```
damassi, line 46: We use screen tracking all over, yet under certain conditions we might not be in a stack or in a tab environment. This is a safe way to invoke hooks yet guard against certain conditions.

src/system/hooks/useTrackScreen.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  const shouldTrackScreen = (() => {
+    // In plain stack-based route
+    if (activeTab == null) {
```
damassi, line 50: `==` is null _or_ undefined

damassi:
@louislecluse - if you scroll through this PR and look for all of the `useTrackScreen("Foo")` calls you can see where each spot will need to be updated in a follow-up.

src/screens/Settings/navigation.tsx
gkartalis: ⭐️

src/system/analytics/AnalyticsProvider.tsx
gkartalis: 🔥 love the implementation ~ way simpler and less verbose 🌟

gkartalis (approved):
awesome!

PR #863, opened 2024-02-02
chore(performance): Add native stack screens, other perf items

damassi:
### Description

This PR brings a nice perf update in: native screens! Just learned about this option in react-navigation today: 

https://reactnavigation.org/docs/native-stack-navigator/

Before we were using JS to transition between screens, but now we're using "real" ios (and android) screens instead, which makes a huge difference. 

Also updated the global tap/active opacity on our UI items on touch, which makes things feel way more responsive too. 

And lastly, memoized our artwork list screens. 

🚗 🌵


src/Navigation.tsx
```diff
@@ -76,7 +76,7 @@ type MiscScreens = {
 
 export type ScreenNames = keyof NavigationScreens
 
-export const StackNav = createStackNavigator<NavigationScreens>()
+export const StackNav = createNativeStackNavigator<NavigationScreens>()
```
damassi, line 79: ❇️ secret sauce is right here

PR #867, opened 2024-02-02
fix(shows): Add preloader state

damassi:
### Description

Fixes an issue where there could be a long hang while we're fetching a list of artist shows, since there's no loading indicator. This adds that to the title area. 

Also fixes a bug where the select mode was enabled on artist -> shows, though there's nothing to select there. Now we disable it if on that tab. 

And finally fixes a bug where select mode (active) is persisted across tab switches.


damassi:
Ugh, select mode tab needs some more work

PR #869, opened 2024-02-02
refactor(select-mode): Refactor select mode action implementation

damassi:
This PR resolves 
- [AMBER-491] 
- [AMBER-493]

### Description

This significantly refactors how we manage select mode interactions in the app, due to a bug I noticed yesterday where the "select mode" button was still visible on the Shows screen, which has no items that one can select. Digging deeper, realized that we were injecting three versions of the buttons into the screen, but outside the normal react lifecycle via a "portal". 

I must have been :zombie: when i wrote that, because that is... quite strange 🙄 

So this updates things by moving select-mode active tab items into the proper location - global state - and then we simply render one single `<SelectModeActions>` component at the top of the react tree screen (ie, on ArtistTabs) and then update our global state whenever a given tab becomes active via 

```tsx
useSetSelectModeActiveTab({
  name: "ShowArtworks",
  items: presentedArtworks,
})
```

From there, we can render the appropriate select mode buttons and everything gets way simpler, while also unlocking some nice behaviors like animated enable / disable / hiding select all, etc. 

https://github.com/artsy/energy/assets/236943/b76e955a-a0b4-498d-946a-03b80f8c7df5

Also fixes the weird artwork -> save to album -> back flash that would happen, and solved by `useMemo`. 

[AMBER-491]: https://artsyproduct.atlassian.net/browse/AMBER-491?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[AMBER-493]: https://artsyproduct.atlassian.net/browse/AMBER-493?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against a375018f2bd6cd864c92d9e4d6e88ca59344ac5d
</p>

jonallured (approved):

PR #874, opened 2024-02-04
feat(artworks): Add more loading indicators, optimize offline images further

damassi:
### Description

This continues work looking at the artwork image presentation across the app, in order to optimize: 
- Adds loading indicators within CachedImage; sometimes it takes a while for those images to return
- Fetches images from Gemini that are appropriate -- a tiny Avatar image doesn't need a 'normalized' 1mb+ size image
- Update the syncManager to point at these new images
- Temporarily adds the Avatar component from palette mobile, which supports image fading on load. Need to [update over there](https://github.com/artsy/palette-mobile/pull/180) after some more discussion.


src/components/Lists/ArtistsList.tsx
```diff
@@ -77,6 +77,7 @@ export const ArtistsList: React.FC<ArtistsListProps> = ({
       numColumns={1}
       maxToRenderPerBatch={20}
       initialNumToRender={20}
+      windowSize={3}
```
damassi, line 80: For large partners, we need to strictly manage how much is loaded up front in the scrollable list. The batches and so on really is no match for my phone / slow internet connection. Lets see how this feels and if it's necessary long-term.

src/screens/Artwork/Artwork.tsx
```diff
@@ -235,7 +233,11 @@ export const Artwork_artworkProps = graphql`
     __typename
     artistNames
     artist {
-      imageUrl
+      image {
+        resized(width: 100, height: 100) {
```
damassi, line 237: Resize it to something tiny, versus the default large image

src/screens/Search/SearchResult.tsx
```diff
@@ -271,7 +266,7 @@ const searchResultQuery = graphql`
             name
             slug
             coverImage {
-              resized(width: $imageSize, version: "normalized") {
+              resized(width: 100, height: 100) {
```
damassi, line 269: Ditto for our search results. Tiny little thumbnails in an avatar image. No need for huge normalized images

PR #876, opened 2024-02-04
feat(performance): Add (online) offline-first cache lookup for images, and only afterwards hit the net

damassi:
### Description

This one is realllly speeds things up if the user has performed a previous offline sync. 

We've got the images stored on disk, why not use those even if the user has an internet connection? So if they're there, grab them from the offline cache _first_ and _then_ fallback to online if not.


src/screens/Settings/DevMenu/CodePushOptions.tsx
```diff
@@ -86,124 +92,136 @@ export const CodePushOptions = () => {
...
-                        )
-                        setLoading(false)
-                      }
+      <Flex
+        flexDirection="row"
+        justifyContent="space-between"
+        alignItems="center"
+      >
+        <Box>
+          {Object.keys(codePushDeploymentKeys).map((deployment) => {
```
damassi, line 102: Light cleanup of codepush dev menu. 

And omg, finally started using CodePush (rather than betas) and it is amazing! 

Thank you so much @brainbicycle. (Related: I'm eager to discuss turning on optimistic (eager) codepushes for partners. The process is so slick / fast.)

PR #882, opened 2024-02-05
fix(installs): Fix install image displays + more darkmode fixes 

damassi:
### Description

- Fixes the install shot displays in lists (didn't respect aspect ratio)
- Appends install shots to saved artworks in images carousel on album list item
- Shows the total count of _all_ items in album, not just images 
- Also removes the "is deleted?" check on album installs because the query is extremely expensive to check all possible combinations of shows. Installs are less subject to change than artworks, so I don't think this will matter much. 
- Darkmode fixes


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 3712485e8216727a09300e00d60aad85724efc7d
</p>

PR #888, opened 2024-02-07
feat(unleash): Add unleash feature flag support 

damassi:
This PR resolves [AMBER-500]

### Description

This adds unleash feature flag support to the app and a new hook, matching our other apps. No support for variants yet, but we can add that when we actually need to use it:

```tsx
import { useFeatureFlag } from "system/hooks/useFeatureFlag" 

const myFeatureFlag = useFeatureFlag("flag_name")
```

cc @artsy/amber-devs 

[AMBER-500]: https://artsyproduct.atlassian.net/browse/AMBER-500?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jpotts244 (approved):
Looks good besides the type check-ers!

damassi:
Ugh that type checking is suspicious. Works fine locally, not on CI. We'll need to keep an eye on it.

